PR #3, opened 2017-03-31
TICK-340 Cleaned array_test

vinther:


tick/base/array_test/src/array_test.cpp
```diff
@@ -27,18 +27,17 @@ double test_constructor_ArrayDouble(unsigned long size){
     return a_i * b_i * c[2] * d[3];
 }
 
-SArrayDoublePtr test_arange(long min, long max) {
+SArrayDoublePtr test_arange(std::int64_t min, std::int64_t max) {
```
Mbompr: long is not authorized? I thought only unsigned long was. Don't we use long in many other places?

tick/base/array_test/src/timer.h
```diff
@@ -6,26 +6,38 @@
 //  Copyright (c) 2015 bacry. All rights reserved.
```
Mbompr: is this still in use?

tick/base/array_test/src/array_test.cpp
vinther: Nope

tick/base/array_test/src/timer.h
vinther: Only in the performance test source. But it was easier to refactor than to make the macros conform to cpplint.

tick/base/array_test/src/array_test.cpp
Mbompr: Oh ok, I'm surprised. Shouldn't we have a typedef just like ulong?

vinther: Yeah, it would be nice to have tick::uint, tick::int, tick::ulong, etc.

tick/base/array_test/src/timer.h
vinther: Yes, I will see if not most of the array_test can go away completely.

PR #6, opened 2017-04-01
TICK-272 Overhaul of the build system for Linux, Mac and Travis

vinther:
Just for Travis


vinther:
Ready

setup.py
```diff
@@ -37,6 +47,22 @@
...
+if python_ver < python_min_ver:
+    txt = 'Python version {0}.{1}.{2} ' \
+          'lower than the required version >= {3}.{4}.{5}.'
+
+    warnings.warn(txt.format(*(python_ver + python_min_ver)))
+
+if sys.platform == 'darwin':
+    from distutils import sysconfig
+    vars = sysconfig.get_config_vars()
+    vars['LDSHARED'] = vars['LDSHARED'].replace('-bundle', '-dynamiclib')
```
Mbompr: Is this to avoid making a bundle? Maybe should it be commented explaining why we should avoid it

setup.py
```diff
@@ -37,6 +47,22 @@
 
 debug_flags = ['DEBUG_COSTLY_THROW']
 
+version_info = sys.version_info
```
Mbompr: does this setup.py script run with python 2? Don't we have the previous syntax error?

setup.py
```diff
@@ -258,25 +282,32 @@ def add_dir_name(dir_name, filenames):
         for opts in [swig_opts, extra_compile_args]:
             opts.extend(["-I%s/" % mod.src, "-I%s/" % mod.swig])
 
-        # Add compile-time location of the dependency library
-        library_dirs.append(os.path.join(mod.build_dir, mod.build))
-
         # Because setuptools produces shared object files with non-standard
```
Mbompr: looks like many lines length are > 80 char

setup.py
```diff
@@ -258,25 +282,32 @@ def add_dir_name(dir_name, filenames):
...
+        # dependencies by using the relative path to the dependency library.
+        if platform.system() == 'Linux':
+            # $ORIGIN refers to the location of the current shared object file
+            # at runtime
+            runtime_library_dirs.append(
+                "\$ORIGIN/%s" % os.path.relpath(mod.build, swig_path.build))
+        else: # Assuming non-Windows builds for now
+            # For OSX builds we use @loader_path instead
+            extra_link_args.append(
+                "-Wl,-rpath,%s" % '@loader_path/%s' % os.path.relpath(mod.build, swig_path.build)
```
Mbompr: `os.path.relpath(mod.build, swig_path.build)` can be given a name (`rel_path_to_shared_object` ?) and shared by osx and linus

setup.py
```diff
@@ -519,21 +540,229 @@ def add_dir_name(dir_name, filenames):
...
 
 
 class CustomInstall(install):
     def run(self):
         self.run_command('build_ext')
-        self.do_egg_install()
+        install.run(self)
+
+
+class TickCommand(Command):
```
Mbompr: For a quicker understanding, we might document it as an abstract class for tick commands

setup.py
```diff
@@ -519,21 +540,229 @@ def add_dir_name(dir_name, filenames):
...
+
+    def initialize_options(self):
+        """Set default values for options."""
+        self.build_jobs = multiprocessing.cpu_count()
+
+    def run(self):
+        relpath = os.path.relpath(self.tick_dir, self.cpp_build_dir)
+        cmake_exe = os.environ.get('TICK_CMAKE', 'cmake')
+
+        from distutils import sysconfig
```
Mbompr: Why is this in a local import?

setup.py
```diff
@@ -519,21 +540,229 @@ def add_dir_name(dir_name, filenames):
...
+            full_path = os.path.abspath(os.path.join(mod.module_ref.build, mod.module_ref.lib_filename))
+
+            cmake_cmd.append(
+                '-DTICKLIB_{}={}'.format(mod.ext_name.upper(),
+                                            full_path))
+
+        os.makedirs(os.path.join(self.cpp_build_dir, 'cpptest'), exist_ok=True)
+        subprocess.check_call(cmake_cmd, cwd=self.cpp_build_dir)
+
+        make_cmd = ['make', 'all', '-j{}'.format(self.build_jobs)]
```
Mbompr: Even if that's not likely to happen, does this parallel build imply that test modules cannot have cross dependencies?

setup.py
```diff
@@ -519,21 +540,229 @@ def add_dir_name(dir_name, filenames):
...
+
+    def run(self):
+        self.run_command('cpptest')
+        self.run_command('pytest')
+
+
+class CleanTick(clean):
+    description = 'cleans all generated and built files'
+
+    def run(self):
```
Mbompr: should'nt we add a `time.sleep(2)` here such that user has 2 seconds to cancel full clean?

setup.py
```diff
@@ -519,21 +540,229 @@ def add_dir_name(dir_name, filenames):
...
+        except ImportError:
+            warnings.warn("Stylecheck by cpplint failed because cpplint "
+                          "is not installed as a Python module")
+
+
+class RunPyLint(TickCommand):
+    # We need to define if and how we run pylint
+
+    description = 'run tick PyLint codestyle check'
+    start_dir = '.'
```
Mbompr: maybe should we output that it is not implemented yet?

tick/CMakeLists.txt
```diff
@@ -36,14 +36,28 @@ if (${GTEST_FOUND})
...
+    find_package(PythonLibs REQUIRED)
 
     include_directories(${GTEST_INCLUDE_DIRS})
 
+    set(TICK_TEST_LIBS
+        ${PYTHON_LIBRARIES}
+        ${GTEST_BOTH_LIBRARIES}
+        ${CMAKE_THREAD_LIBS_INIT})
+
     if(APPLE)
```
Mbompr: this is necessary for Apple as otherwise everything is included in the docker?

tick/base/array/tests/src/CMakeLists.txt
```diff
@@ -1,5 +1,14 @@
 add_executable(tick_test_array array_gtest.cpp)
 add_executable(tick_test_varray varray_gtest.cpp)
 
-target_link_libraries(tick_test_array ${GTEST_BOTH_LIBRARIES} ${CMAKE_THREAD_LIBS_INIT} ${EXTRA_TEST_LIBS} array)
-target_link_libraries(tick_test_varray ${GTEST_BOTH_LIBRARIES} ${CMAKE_THREAD_LIBS_INIT} ${EXTRA_TEST_LIBS} array)
+set(CMAKE_VERBOSE_MAKEFILE ON)
```
Mbompr: is it left on purpose?

.travis.yml
```diff
@@ -1,44 +1,50 @@
...
+    - $HOME/googletest
 
 matrix:
   include:
   - os: linux
-    language: generic
-    env: TOXENV=py34
+    services: docker
+    env:
+      - DOCKER_IMAGE=vinther/tick_manylinux1_x86_64:1.2
```
vinther: No, tick_manylinux1_x86_64 and tick_ubuntu are just images where tick can be built without the need to install dependencies. The dependencies are already installed in the images.

setup.py
vinther: Nope, only the compilation stage is parallelized

tick/CMakeLists.txt
vinther: No, docker does not play in here. It's only because OSX handles rpaths a bit different than linux

.travis.yml
```diff
@@ -1,44 +1,50 @@
...
   only:
     - master
 
+# Cache is irrelevant for linux/docker builds, but useful to speed up osx
 cache:
   directories:
     - $HOME/.pyenv
-    - googletest
-    - swig-3.0.10
+    - $HOME/googletest
```
vinther: There's no caching of brew directories. It _should_ be faster to get deps via brew than the Travis cache

PR #8, opened 2017-04-11
Tick 294 sum gaussian

achab:
Implementation of the EM algorithm (https://arxiv.org/pdf/1602.04511v2.pdf) where Hawkes kernels are expressed as sum of gaussian functions.

### DIFF
--- a/tick/inference/hawkes_sumgaussians.py
+++ b/tick/inference/hawkes_sumgaussians.py
@@ -1,0 +1,404 @@
+import math
+import numpy as np
+from scipy.stats import norm
+from tick.inference.base import LearnerHawkesNoParam
+from tick.optim.solver.base.utils import relative_distance
+from .build.inference import HawkesSumGaussians as _HawkesSumGaussians
+
+
+class HawkesSumGaussians(LearnerHawkesNoParam):
+    """A class that implements parametric inference for Hawkes processes
+    with parametrisation of the kernels as sum of Gaussian basis functions
+    and a mix of Lasso and group-lasso regularization
+
+    Hawkes processes are point processes defined by the intensity:
+
+    .. math::
+        \\forall i \\in [1 \\dots D], \\quad
+        \\lambda_i(t) = \\mu_i + \\sum_{j=1}^D
+        \\sum_{t_k^j < t} \\phi_{ij}(t - t_k^j)
+
+    where
+
+    * :math:`D` is the number of nodes
+    * :math:`\mu_i` are the baseline intensities
+    * :math:`\phi_{ij}` are the kernels
+    * :math:`t_k^j` are the timestamps of all events of node :math:`j`
+
+    and with an parametrisation of the kernels as sum of Gaussian basis 
+    functions
+
+    .. math::
+        \phi_{ij}(t) = \sum_{m=1}^M \\alpha^{ij}_m f (t - t_m), \\quad
+        f(t) = (2 \\pi \\sigma^2)^{-1} \exp(- t^2 / (2 \\sigma^2))
+
+    In our implementation we denote:
+
+    * Integer :math:`D` by the attribute `n_nodes`
+    * Vector :math:`\mu \in \mathbb{R}^{D}` by the attribute
+      `baseline`
+    * Vector :math:`(t_m) \in \mathbb{R}^{M}` by the variable
+      `means_gaussians`
+    * Number :math:`\\sigma` by the variable `std_gaussian`
+    * Tensor 
+      :math:`A = (\\alpha^{ij}_m)_{ijm} \in \mathbb{R}^{D \\times D \\times M}`
+      by the attribute `amplitudes`
+
+    Parameters
+    ----------
+    max_mean_gaussian : `float`
+        The mean of the last Gaussian basis function. This can be considered
+        a proxy of the kernel support.
+
+    n_gaussians : `int`
+        The number of Gaussian basis functions used to approximate each kernel.
+
+    step_size : `float`
+        The step-size used in the optimization for the EM algorithm.
+
+    C : `float`, default=1e3
+        Level of penalization
+
+    lasso_grouplasso_ratio : `float`, default=0.5
+        Ratio of Lasso-Nuclear regularization mixing parameter with
+        0 <= ratio <= 1.
+
+        * For ratio = 0 this is Group-Lasso regularization
+        * For ratio = 1 this is lasso (L1) regularization
+        * For 0 < ratio < 1, the regularization is a linear combination
+          of Lasso and Group-Lasso.
+
+    max_iter : `int`, default=50
+        Maximum number of iterations of the solving algorithm
+
+    tol : `float`, default=1e-5
+        The tolerance of the solving algorithm (iterations stop when the
+        stopping criterion is below it). If not reached it does ``max_iter``
+        iterations
+
+    n_threads : `int`, default=1
+        Number of threads used for parallel computation.
+
+    verbose : `bool`, default=False
+        If `True`, we verbose things
+
+        * if `int <= 0`: the number of physical cores available on the CPU
+        * otherwise the desired number of threads
+
+    print_every : `int`, default=10
+        Print history information when ``n_iter`` (iteration number) is
+        a multiple of ``print_every``
+
+    record_every : `int`, default=10
+        Record history information when ``n_iter`` (iteration number) is
+        a multiple of ``record_every``
+
+    Other Parameters
+    ----------------
+    approx : `int`, default=0 (read-only)
+        Level of approximation used for computing exponential functions
+
+        * if 0: no approximation
+        * if 1: a fast approximated exponential function is used
+
+    em_max_iter : `int`, default=30
+        Maximum number of loop for inner em algorithm.
+
+    em_tol : `float`, default=None
+        Tolerance of loop for inner em algorithm. If relative difference of
+        baseline and adjacency goes bellow this tolerance, em inner loop
+        will stop.
+        If None, it will be set given a heuristic which look at last
+
+    Attributes
+    ----------
+    n_nodes : `int`
+        Number of nodes / components in the Hawkes model
+        
+    baseline : `np.array`, shape=(n_nodes,)
+        Inferred baseline of each component's intensity
+
+    amplitudes : `np.ndarray`, shape=(n_nodes, n_nodes, n_gaussians)
+        Inferred adjacency matrix
+        
+    means_gaussians : `np.array`, shape=(n_gaussians,)
+        The means of the Gaussian basis functions.
+
+    std_gaussian : `float`
+        The standard deviation of each Gaussian basis function.
+
+    References
+    ----------
+    Xu, Farajtabar, and Zha (2016, June) in ICML
+    `Learning Granger Causality for Hawkes Processes`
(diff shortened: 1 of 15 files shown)


tick/inference/hawkes_sumgaussians.py
```diff
@@ -0,0 +1,404 @@
...
+    * Vector :math:`\mu \in \mathbb{R}^{D}` by the attribute
+      `baseline`
+    * Vector :math:`(t_m) \in \mathbb{R}^{M}` by the variable
+      `means_gaussians`
+    * Number :math:`\\sigma` by the variable `std_gaussian`
+    * Tensor 
+      :math:`A = (\\alpha^{ij}_m)_{ijm} \in \mathbb{R}^{D \\times D \\times M}`
+      by the attribute `amplitudes`
+    * Number :math:`\\beta \in \mathbb{R}` by the parameter `decay`. This
+      parameter is given to the model
```
Mbompr: it would be cool if `means_gaussians` and `std_gaussian` was attributes of the model. It could be done with a property.

tick/inference/src/hawkes_sumgaussians.h
```diff
@@ -0,0 +1,99 @@
...
+  //! @brief Standard deviation of the gaussian functions. This bandlimit
+  //! is computed using Silverman's rule of thumb.
+  double std_gaussian;
+
+  //! @brief Useful constants that appear in weights computation.
+  double std_gaussian_sq = std_gaussian*std_gaussian;
+  double norm_constant_gauss = std_gaussian * std::sqrt(2.*M_PI);
+  double norm_constant_erf = std_gaussian * std::sqrt(2);
+
+  //! @brief Step size used in update formules (7) and (8)
```
Mbompr: formulas

tick/inference/hawkes_sumgaussians.py
```diff
@@ -0,0 +1,404 @@
...
+        },
+        "_C": {"writable": False},
+        "baseline": {"writable": False},
+        "amplitudes": {"writable": False},
+        "_prox_l1": {"writable": False},
+        "_lasso_grouplasso_ratio": {"writable": False},
+        "approx": {"writable": False}
+    }
+
+    def __init__(self, n_gaussians, max_mean_gaussian=5, step_size=1e-7,
```
Mbompr: I think we can have a default value to `n_gaussians` (it is almost like a kernel_size in EM) but I find that `max_mean_gaussian`'s default value (5) is very arbitrary and hence we should force the user to set it manually.

tick/inference/src/hawkes_sumgaussians.h
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * \class HawkesSumGaussians
+ * \brief Class for implementation of the algorithm described in the paper
+ * `Learning Granger Causality for Hawkes Processes`
+ *  by Xu, Farajtabar, and Zha (2016) in ICML
+ */
+class HawkesSumGaussians : public ModelHawkesList {
+  //! @brief Number of gaussian functions to approximate each kernel
+  int n_gaussians;
```
Mbompr: type should be either int or ulong but your switching too often :), same for em_max_iter.

tick/inference/src/hawkes_sumgaussians.cpp
```diff
@@ -0,0 +1,332 @@
...
+  this->em_max_iter = em_max_iter;
+}
+
+double HawkesSumGaussians::get_max_mean_gaussian() const {
+  return max_mean_gaussian;
+}
+
+void HawkesSumGaussians::set_max_mean_gaussian(const double max_mean_gaussian) {
+  if (max_mean_gaussian <= 0) {
+    TICK_ERROR("max_mean_gaussian must be positive, received " << max_mean_gaussian);
```
Mbompr: Shouldn't this set `weights_computed = false;` ?

tick/inference/src/hawkes_sumgaussians.cpp
```diff
@@ -0,0 +1,332 @@
...
+      diff_norm += tmp * tmp;
+    }
+    diff_norm = std::sqrt(diff_norm);
+    // if condition is met set a^{(k+1)}_{uv} to zero, else use the thresholding
+    // operator (see Eq. 8 of the paper)
+    if (diff_norm <= step_size * penalty_param_grouplasso) {
+      for (ulong m = 0; m < n_gaussians; m++) {
+        amplitudes_u[v * n_gaussians + m] = 0.;
+      }
+      //amplitudes_uv.fill(0.);
```
Mbompr: left comment

tick/inference/hawkes_sumgaussians.py
```diff
@@ -0,0 +1,404 @@
...
+        if val < 0 or val is None:
+            raise ValueError("`C` must be positive, got %s" % str(val))
+        else:
+            self._set("_C", val)
+
+    @property
+    def lasso_grouplasso_ratio(self):
+        param_lasso = self._learner.get_penalty_param_lasso()
+        param_grouplasso = self._learner.get_penalty_param_grouplasso()
+        if param_lasso is None or param_grouplasso is None:
```
Mbompr: When does this happen?

Mbompr: I a bit frightened it could cause a silent unwanted behavior

tick/inference/hawkes_sumgaussians.py
```diff
@@ -0,0 +1,404 @@
...
+    def strength_lasso(self):
+        return self.lasso_grouplasso_ratio / self.C
+
+    @strength_lasso.setter
+    def strength_lasso(self, val):
+        if val < 0:
+            raise ValueError("`strength_lasso` must be positive, "
+                             "got {}".format(val))
+        else:
+            self._set("penalty_param_lasso")
```
Mbompr: Something must be missing here

tick/inference/hawkes_sumgaussians.py
```diff
@@ -0,0 +1,404 @@
...
+        """
+        x = np.zeros_like(abscissa_array)
+        bandlimit = self.max_mean_gaussian / (self.n_gaussians * math.pi)
+        for m in range(self.n_gaussians):
+            t_m = m * self.max_mean_gaussian * 1. / self.n_gaussians
+            x += self.amplitudes[i,j,m]*norm.pdf((abscissa_array-t_m)/bandlimit)/bandlimit
+        return x
+
+    def objective(self, coeffs, loss: float = None):
+        raise NotImplementedError()
```
Mbompr: would be cool to add a kernel norms computations if it's not too complicated.

tick/inference/tests/hawkes_sumgaussians.py
```diff
@@ -0,0 +1,112 @@
...
+        self.assertEqual(learner.max_mean_gaussian, self.float_2)
+        self.assertEqual(learner._learner.get_max_mean_gaussian(), self.float_2)
+
+        learner = HawkesSumGaussians(1, step_size=self.float_1)
+        self.assertEqual(learner.step_size, self.float_1)
+        self.assertEqual(learner._learner.get_step_size(), self.float_1)
+        learner.step_size = self.float_2
+        self.assertEqual(learner.step_size, self.float_2)
+        self.assertEqual(learner._learner.get_step_size(), self.float_2)
+
```
Mbompr: you should test here you penalty parameters (as they are not straight forward)

tick/inference/hawkes_sumgaussians.py
achab: done

tick/inference/hawkes_sumgaussians.py
achab: indeed, I fixed it

tick/inference/hawkes_sumgaussians.py
achab: ok

tick/inference/hawkes_sumgaussians.py
achab: indeed

tick/inference/src/hawkes_sumgaussians.h
achab: you're right, no place for hesitation here

tick/inference/src/hawkes_sumgaussians.cpp
achab: most def

tick/inference/src/hawkes_sumgaussians.h
achab: -1 pt for wrong translation

tick/inference/hawkes_sumgaussians.py
achab: I should add a new function ? It can be done easily

tick/inference/src/hawkes_sumgaussians.cpp
achab: removed

tick/inference/hawkes_sumgaussians.py
```diff
@@ -120,6 +120,16 @@ class HawkesSumGaussians(LearnerHawkesNoParam):
     n_gaussians : `int`
         Number of gaussians used to approximate each kernel
 
+    max_mean_gaussian : `float`
+        The mean of the last Gaussian basis function. This can be considered
+        a proxy of the kernel support.
+
+    means_gaussians : `np.array`, shape=(n_gaussians,)
```
Mbompr: This should be in the attributes sections

tick/inference/hawkes_sumgaussians.py
```diff
@@ -120,6 +120,16 @@ class HawkesSumGaussians(LearnerHawkesNoParam):
     n_gaussians : `int`
         Number of gaussians used to approximate each kernel
 
+    max_mean_gaussian : `float`
```
Mbompr: Not very important but it's always better if parameters documentations are in the same order as in the init function :)

Mbompr:
Don't forget to make it cpplint compliant (cf https://github.com/X-DataInitiative/tick#pull-requests)

tick/inference/tests/hawkes_sumgaussians.py
Mbompr: I still think it would be a good idea :).

Testing means_gaussians and std_gaussians as well

tick/inference/hawkes_sumgaussians.py
```diff
@@ -0,0 +1,431 @@
...
+            1d array containing all the times at which this kernel will
+            computes it value
+
+        Returns
+        -------
+        output : `np.ndarray`, shape=(n_points, )
+            1d array containing the values of the specified kernels at the
+            given times.
+        """
+        x = np.zeros_like(abscissa_array)
```
Mbompr: maybe use self.means_gaussians and self.std_gaussians here?

tick/inference/hawkes_sumgaussians.py
Mbompr: Can't see it, am I missing something?

tick/inference/tests/hawkes_sumgaussians.py
achab: done

tick/inference/hawkes_sumgaussians.py
achab: indeed, more clear

PR #10, opened 2017-04-25
TICK-347 add google analytics on documentation

Mbompr:


doc/conf.py
```diff
@@ -219,6 +217,7 @@
     'css_files': ['_static/css/custom.css'],
 }
 
+googleanalytics_id = 'UA-97947441-1'
```
vinther: This id is not secret in any way?

doc/conf.py
```diff
@@ -56,6 +53,7 @@
     'matplotlib.sphinxext.plot_directive',
     'sphinxext.gen_rst',
     'sphinxext.directives',
+    'sphinxext.google_analytics'
```
vinther: Is this plugin provided with sphinx by default?

doc/conf.py
Mbompr: No, at least you cannot hide it...
https://webapps.stackexchange.com/questions/5440/is-it-possible-to-hide-a-google-analytics-user-id-from-source
I have also enable "hacking defense"
https://blog.kissmetrics.com/protect-analytics-from-hacking/

doc/conf.py
Mbompr: no, this it the file I have added manually. Otherwise we need to add https://pypi.python.org/pypi/sphinxcontrib-googleanalytics/ as a requirement

PR #12, opened 2017-05-02
Tick 351 simulate hawkes with non consta

Mbompr:


Mbompr:
First commit is only code refactoring

tick/simulation/src/hawkes.h
```diff
@@ -115,24 +75,41 @@ class Hawkes : public PP {
...
-  void set_mu(unsigned int i, const HawkesMuPtr &mu);
+  void set_baseline(unsigned int i, double baseline);
+
+  /**
+   * @brief Set baseline for a specific dimension
+   * \param i : the dimension
+   * \param time_function : a TimeFunction that will be used to construct a
+   * HawkesTimeFunctionBaseline
+   */
+  void set_baseline(unsigned int i, TimeFunction time_function);
```
vinther: Can be const-ref?

tick/simulation/src/hawkes.h
```diff
@@ -115,24 +75,41 @@ class Hawkes : public PP {
...
+
+  /**
+  * @brief Set baseline for a specific dimension
+  * \param i : the dimension
+  * \param times : times that will be used to construct a HawkesTimeFunctionBaseline
+  * \param values : values that will be used to construct a HawkesTimeFunctionBaseline
+  */
+  void set_baseline(unsigned int i, ArrayDouble &times, ArrayDouble &values);
```
vinther: Can be const-ref?

Mbompr: Can't be as array[i] is not const

PR #16, opened 2017-05-10
TICK-356 Add ProxEquality

stephanegaiffas:
Create a ProxEquality that imposes an equality constraint
This is stupid but very useful for Maryan and the CNAM project 
(in order to refit after TV regularization)


stephanegaiffas:
@MaryanMorel : ready for review. 
@Mbompr : Can Maryan merge in master ?

PR #17, opened 2017-05-12
TICK-357 Add sparse longitudinal feature product.

MaryanMorel:
Serialization of the C++ object does not seems to work yet, thus the preprocessing for sparse infinite exposure is not parallel yet.

### DIFF
--- a/tick/preprocessing/longitudinal_features_product.py
+++ b/tick/preprocessing/longitudinal_features_product.py
@@ -1,0 +1,276 @@
+import numpy as np
+import scipy.sparse as sps
+from itertools import combinations
+from copy import deepcopy
+from scipy.misc import comb
+from sklearn.externals.joblib import Parallel, delayed
+from tick.base import Base
+from tick.inference.base.learner_optim import LearnerOptim
+from .build.preprocessing import SparseLongitudinalFeaturesProduct
+
+
+class LongitudinalFeaturesProduct(Base):
+    """Transforms longitudinal exposure features to add the corresponding
+    product features. 
+
+    This preprocessor transform an input list of `n_samples` numpy arrays or 
+    csr_matrices of shape `(n_intervals, n_features)` so as to add columns
+    representing the product of combination of two features. It outputs a list
+    of `n_samples` numpy arrays or  csr_matrices of shape `(n_intervals, 
+    n_features + comb(n_features, 2))`.
+
+    Exposure can take two forms:
+    - short repeated exposures: in that case, each column of the numpy arrays 
+    or csr matrices can contain multiple ones, each one representing an exposure
+    for a particular time bucket.
+    - infinite unique exposures: in that case, each column of the numpy arrays
+    or csr matrices can only contain a single one, corresponding to the starting
+    date of the exposure.
+
+    Parameters
+    ----------
+    exposure_type : {'infinite', 'short'}, default='infinite'
+        Either 'infinite' for infinite unique exposures or 'short' for short
+        repeated exposures.
+
+    n_threads : `int`, default=-1
+        Number of tasks to run in parallel. If set to -1, the number of tasks is 
+        set to the number of cores.
+
+    Attributes
+    ----------
+    mapper : `dict`
+        Map product features to column indexes of the resulting matrices.
+
+    Examples
+    --------
+    >>> from pprint import pprint
+    >>> from scipy.sparse import csr_matrix
+    >>> from tick.preprocessing.longitudinal_features_product import LongitudinalFeaturesProduct
+    >>> infinite_exposures = [csr_matrix([[0, 1, 0],
+    ...                                   [0, 0, 0],
+    ...                                   [0, 0, 1]], dtype="float64"),
+    ...                       csr_matrix([[1, 1, 0],
+    ...                                   [0, 0, 1],
+    ...                                   [0, 0, 0]], dtype="float64")
+    ...                       ]
+    >>> lfp = LongitudinalFeaturesProduct(exposure_type="infinite")
+    >>> product_features = lfp.fit_transform(infinite_exposures)
+    >>> # output comes as a list of sparse matrices or 2D numpy arrays
+    >>> product_features.__class__
+    <class 'list'>
+    >>> pprint([x.toarray() for x in product_features])
+    [array([[ 0.,  1.,  0.,  0.,  0.,  0.],
+           [ 0.,  0.,  0.,  0.,  0.,  0.],
+           [ 0.,  0.,  1.,  0.,  0.,  1.]]),
+     array([[ 1.,  1.,  0.,  1.,  0.,  0.],
+           [ 0.,  0.,  1.,  0.,  1.,  1.],
+           [ 0.,  0.,  0.,  0.,  0.,  0.]])]
+    """
+
+    _attrinfos = {
+        "exposure_type": {"writable": False},
+        "_mapper": {"writable": False},
+        "_n_init_features": {"writable": False},
+        "_n_output_features": {"writable": False},
+        "_n_intervals": {"writable": False},
+        "_preprocessor": {"writable": False},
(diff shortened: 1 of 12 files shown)


tick/preprocessing/longitudinal_features_product.py
```diff
@@ -0,0 +1,236 @@
...
+
+from tick.base import Base
+from itertools import combinations
+import scipy.sparse as sps
+from scipy.misc import comb
+from sklearn.externals.joblib import Parallel, delayed
+from .build.preprocessing import SparseLongitudinalFeaturesProduct
+
+
+class LongitudinalFeaturesProduct(Base):  # or LongitudinalExposureProduct ?
```
stephanegaiffas: LongitudinalFeaturesProduct is good

tick/preprocessing/longitudinal_features_product.py
```diff
@@ -0,0 +1,236 @@
...
+    of `n_samples` numpy arrays or  csr_matrices of shape `(n_intervals, 
+    n_features + comb(n_features, 2))`.
+    
+    Exposure can take two forms:
+    - short repeated exposures: in that case, each column of the numpy arrays 
+    or csr matrices can contain multiple ones, each one representing an exposure
+    for a particular time bucket.
+    - infinite unique exposures: in that case, each column of the numpy arrays
+    or csr matrices can only contain a single one, corresponding to the starting
+     date of the exposure.
```
stephanegaiffas: Now, that a precise doctoring !

tick/preprocessing/longitudinal_features_product.py
```diff
@@ -0,0 +1,236 @@
...
+
+    Parameters
+    ----------
+    exposure_type : `string`, default="infinite"
```
stephanegaiffas: Please write instead

exposure_type : {'infinite', 'short'}, default='infinite'

tick/preprocessing/longitudinal_features_product.py
```diff
@@ -0,0 +1,236 @@
...
+        Either "infinite" for infinite unique exposures or "short" for short
```
stephanegaiffas: And single quotes :)

tick/preprocessing/longitudinal_features_product.py
```diff
@@ -0,0 +1,236 @@
...
+        repeated exposures.
+
```
stephanegaiffas: docstring for n_jobs is missing. 
Maybe this should be called n_threads ? (If I recall well, we use n_threads elsewhere, not n_jobs ?

tick/preprocessing/longitudinal_features_product.py
```diff
@@ -0,0 +1,236 @@
...
+    def __init__(self, exposure_type="infinite", n_jobs=-1):
+        Base.__init__(self)
+        if exposure_type not in ["infinite", "short"]:
+            raise ValueError("exposure_type should be either 'infinite' or\
+             'short', not %s" %exposure_type)
+        self.n_jobs = n_jobs
+        self.exposure_type = exposure_type
+        self.reset()
+
+    def reset(self):
```
stephanegaiffas: missing doctoring. What is it used for ?

tick/preprocessing/longitudinal_features_product.py
```diff
@@ -0,0 +1,236 @@
...
+        """Get the mapping between the feature products and column indexes.
+
+        Returns
+        -------
+        output : `dict`
+            The column index - feature mapping.
+        """
+        if not self._fitted:
+            raise ValueError("cannot get mapper if object has not been fitted.")
+        return self._mapper.copy()
```
stephanegaiffas: I think that copy won't do a copy of the nested values of the dict. 
If necessary, use deepcopy.

tick/preprocessing/longitudinal_features_product.py
```diff
@@ -0,0 +1,236 @@
...
+
+    def fit(self, X):
+        """Fit the feature product using the features matrices list.
+
+        Parameters
+        ----------
+        X : `[np.ndarrays]`  or `[csr_matrices]`, shape=(n_intervals, n_features)
```
stephanegaiffas: `list` of `np.ndarray` or `list` of `scipy.sparse.csr_matrix`

tick/preprocessing/longitudinal_features_product.py
```diff
@@ -0,0 +1,236 @@
...
+            The list of features matrices.
+
+        Returns
+        -------
+        output : `LongitudinalFeaturesProduct`
+            The fitted current instance.
+        """
+        self.reset()
+        # TODO: check input type ?
```
stephanegaiffas: Definitively. You need to check that all shapes are equal and use _safe_array from our inference tools

tick/preprocessing/longitudinal_features_product.py
```diff
@@ -0,0 +1,236 @@
...
+        # TODO: Check that all the matrices have the same size ?
+        n_intervals, n_init_features = X[0].shape
+        if n_init_features < 2:
+            raise ValueError("There should at least two features to compute\
```
stephanegaiffas: There should ``be`` at least

tick/preprocessing/longitudinal_features_product.py
```diff
@@ -0,0 +1,236 @@
...
+        """Add the product features to the given features matrices list.
+
+        Parameters
+        ----------
+        X : `[np.ndarrays]`  or `[csr_matrices]`, shape=(n_intervals, n_features)
+            The list of features matrices.
+
+        Returns
+        -------
+        output : `[np.ndarrays]`  or `[csr_matrices]`, shape=(n_intervals, n_features)
```
stephanegaiffas: Same as above

tick/preprocessing/longitudinal_features_product.py
```diff
@@ -0,0 +1,236 @@
...
+        matrices list.
+
+        Parameters
+        ----------
+        X : `[np.ndarrays]`  or `[csr_matrices]`, shape=(n_intervals, n_features)
+            The list of features matrices.
+
+        Returns
+        -------
+        output : `pd.DataFrame`
```
stephanegaiffas: It's not a DataFrame ???

tick/preprocessing/longitudinal_features_product.py
```diff
@@ -0,0 +1,236 @@
...
+        -------
+        output : `[np.ndarrays]`  or `[csr_matrices]`, shape=(n_intervals, n_features)
+            The list of features matrices with added product features.
+        """
+
+        # TODO: check matrix dimensions and consistency with self.fit ?
+        if self.exposure_type == "short":
+            X_with_products = self.short_exposure_products(X)
+        elif self.exposure_type == "infinite":
+            X_with_products = self.infinite_exposure_products(X)
```
stephanegaiffas: else raise a ValueError or something like that (in case a user plays with set._set()

tick/preprocessing/longitudinal_features_product.py
```diff
@@ -0,0 +1,236 @@
...
+            binarized.
+        """
+        self.fit(X)
+        X_with_products = self.transform(X)
+
+        return X_with_products
+
+    def infinite_exposure_products(self, X):
+        if sps.issparse(X[0]):
+            X_with_products = [self._sparse_infinite_product(arr) for arr in X]  # TODO: fix multiprocessing
```
stephanegaiffas: **Remark**
This loop should be done in C++, but we can keep it in python for now.
Multiprocessing should and could be done from C++ as well... but later also :)
In general, maybe many things done here can be done in C++ for efficiency

MaryanMorel: For this loop to be done in C++, we should be able to build sparse matrices directly in C++. Do we have the tools to do that ?

MaryanMorel:
Added the preprocessor to the doc

PR #21, opened 2017-05-21
TICK-371 Adds an example comparing tick and scikit-learn for L1 logistic regression

stephanegaiffas:


stephanegaiffas:
@Mbompr : so do you approve ?

examples/plot_logistic_tick_vs_scikit.py
```diff
@@ -0,0 +1,84 @@
...
+from sklearn.linear_model import LogisticRegression as LogRegScikit
+
+from tick.dataset import fetch_tick_dataset
+from tick.inference import LogisticRegression as LogRegTick
+
+train_set = fetch_tick_dataset('binary/adult/adult.trn.bz2')
+test_set = fetch_tick_dataset('binary/adult/adult.tst.bz2')
+
+clf_tick = LogRegTick(C=1e5, penalty='l1')
+clf_scikit = LogRegScikit(penalty='l1')
```
Mbompr: Do you think we should add a shared tol parameter?

stephanegaiffas: Yes I'll do it.

examples/plot_logistic_tick_vs_scikit.py
```diff
@@ -0,0 +1,84 @@
...
+
+t1 = time()
+clf_tick.fit(train_set[0], train_set[1])
+t_tick = time() - t1
```
Mbompr: for performance measurement it's better to use time.clock()
http://pythoncentral.io/measure-time-in-python-time-time-vs-time-clock/

PR #22, opened 2017-05-22
TICK-359 SCCS Simulation

MaryanMorel:
Add SCCS simulation class

### DIFF
--- a/doc/modules/simulation.rst
+++ b/doc/modules/simulation.rst
@@ -138,6 +138,9 @@ Poisson regression (exponential link)  :math:`\text{Poisson}(e^{w^\top x + b})` 
 3. Survival analysis simulation
 ===============================
 
+3.1 Cox regression simulation (proportional hazards)
+----------------------------------------------------
+
 We provide a class for the simulation of a Cox regression model with right-censoring.
 This generates data in the form of i.i.d triplets :math:`(x_i, t_i, c_i)`
 for :math:`i=1, \ldots, n`, where :math:`x_i \in \mathbb R^d` is a features vector,
@@ -163,6 +166,18 @@ Cox regression with right-censoring  :class:`tick.simulation.SimuCoxReg`
 
 .. plot:: ../examples/plot_simulation_coxreg.py
     :include-source:
+
+3.2 Self-controlled case series (SCCS)
+--------------------------------------
+
+We provide a class for the simulation of a SCCS regression model with right-censoring.
+
+.. autosummary::
+   :toctree: generated/
+   :template: class.rst
+
+   simulation.SimuSCCS
+
 
 
 4. Point process simulation
--- a/tick/preprocessing/longitudinal_features_lagger.py
+++ b/tick/preprocessing/longitudinal_features_lagger.py
@@ -48,7 +48,7 @@ class LongitudinalFeaturesLagger(Base):
     ...                         [0, 0, 1],
     ...                         [0, 0, 0]], dtype="float64")
     ...             ]
-    >>> censoring = np.array([3, 2], dtype="uint64").reshape((-1, 1))
+    >>> censoring = np.array([3, 2], dtype="uint64")
     >>> lfl = LongitudinalFeaturesLagger(n_lags=2)
     >>> product_features = lfl.fit_transform(features)
     >>> # output comes as a list of sparse matrices or 2D numpy arrays
@@ -114,7 +114,7 @@ class LongitudinalFeaturesLagger(Base):
             shape=(n_intervals, n_features)
             The list of features matrices.
             
-        censoring : `numpy.ndarray`, shape=(n_samples, 1), dtype="uint64"
+        censoring : `numpy.ndarray`, shape=(n_samples,), dtype="uint64"
             The censoring data. This array should contain integers in 
             [1, n_intervals]. If the value i is equal to n_intervals, then there
             is no censoring for sample i. If censoring = c < n_intervals, then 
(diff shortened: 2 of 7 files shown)


tick/simulation/SCCS.py
```diff
@@ -0,0 +1,379 @@
...
+        if self.distribution == "poisson":
+            y = self._simulate_outcome_from_poisson(features, self.coeffs,
+                                                    self.first_tick_only)
+        else:
+            y = self._simulate_outcome_from_multi(features, self.coeffs,)
+        return y
+
+    @staticmethod
+    def _simulate_outcome_from_multi(features, coeffs):
+        inner_product = [f.dot(coeffs) for f in features]
```
stephanegaiffas: dot_products

tick/simulation/SCCS.py
```diff
@@ -0,0 +1,379 @@
...
+
+        def simulate(inner_prod):
+            inner_prod -= inner_prod.max()
+            probabilities = np.exp(inner_prod) / \
+                            np.sum(np.exp(inner_prod))
+            y = np.random.multinomial(1, probabilities)
+            return y.astype("uint64")
+
+        return [simulate(i) for i in inner_product]
```
stephanegaiffas:
```python
for dot_product in dot_products
```
i is really misleading

tick/simulation/SCCS.py
```diff
@@ -0,0 +1,379 @@
...
+
+    @staticmethod
+    def _simulate_outcome_from_poisson(features, coeffs, first_tick_only=True):
+        inner_product = [f.dot(coeffs) for f in features]
```
stephanegaiffas: same as above here

tick/simulation/SCCS.py
```diff
@@ -0,0 +1,379 @@
+"""AR-SCCS simulations utility."""
```
stephanegaiffas: please use lowercase in filenames only

tick/simulation/SCCS.py
```diff
@@ -0,0 +1,379 @@
...
+
+        return [simulate(i) for i in inner_product]
+
+    @staticmethod
+    def _simulate_outcome_from_poisson(features, coeffs, first_tick_only=True):
+        inner_product = [f.dot(coeffs) for f in features]
+
+        def simulate(inner_prod):
+            intercept = -inner_prod.max()
+            intensities = np.exp(intercept + inner_prod)
```
stephanegaiffas: What's the use of intercept, since it's put back right after ?

tick/simulation/SCCS.py
```diff
@@ -0,0 +1,379 @@
...
+            ticks = np.random.poisson(lam=intensities)
```
stephanegaiffas: intensities.max() should not be too large... otherwise raise a ValueError saying that intensity is too large and Poisson simulation will fail because of overflows

tick/simulation/tests/SCCS_test.py
```diff
@@ -0,0 +1,79 @@
...
+            self.assertEqual(coeffs.shape, (n_features * (n_lags + 1),))
+
+        n_samples = [1, 100]
+        n_features = [1, 3]
+        sparse = [True, False]
+        exposure_type = ["infinite", "short"]
+        distribution = ["multinomial", "poisson"]
+        first_tick_only = [True, False]
+        censoring = [True, False]
+        for n in n_samples:
```
stephanegaiffas: Ouch, this hurts the eye ! itertools.product is your friend : 
```python
prods = itertools.product(n_features, exposure_type, distribution, first_tick_only, censoring)
for n_feat, e, d, f, c in prods:
    pass
```

tick/simulation/SCCS.py
```diff
@@ -0,0 +1,379 @@
...
+        
+    censoring_prob : `float`, default=.7
+        Probability that a sample is censored. Should be in [0, 1].
+        
+    censoring_intensity : `float`, default=.9
+        The number of censored time intervals are drawn from a Poisson 
+        distribution with intensity equal to `censoring_intensity`. The higher,
+        the more intervals will be censored. Should be greater than 0.
+        
+    coeffs : `numpy.ndarray` of shape (n_features * (n_lags + 1)), default=None
```
MaryanMorel: coeffs shape depends on n_lags and n_features, so it seemed weird to have it as the first argument. The default value of coeffs is None. In that case, random coefficients are generated. 
I can indeed add a function to generate more "structured" random coefficients

tick/simulation/SCCS.py
```diff
@@ -0,0 +1,379 @@
...
+            [ 1.,  1.]])
+    ]
+    >>> print(labels)
+    [array([0, 0, 1], dtype=uint64), array([0, 0, 1], dtype=uint64)]
+    >>> print(censoring)
+    [3 3]
+    >>> print(coeffs)
+    [ 0.54738557 -0.15109073  0.71345739  1.67633284 -0.25656871 -0.25655065]
+     """
+    # TODO: case where censoring_intensity is too high regarding n_intervals
```
MaryanMorel: It's wip ;)

tick/simulation/SCCS.py
MaryanMorel: `intercept` is just here to avoid overflows. I subtract the max from the intensity as we do not estimate the baseline with SCCS. If you think it is a problem, I can add a parameter to deactivate this regularization.

tick/simulation/SCCS.py
MaryanMorel: Cf. comment about the intercept

tick/simulation/tests/SCCS_test.py
MaryanMorel: I didn't knew that one, It's indeed way better

PR #24, opened 2017-05-25
TICK-362 Added extra tests to set_model and set_prox and test_set_mod…

stephanegaiffas:
Added extra tests to set_model and set_prox and test_set_model_and_set_prox in TestSolver unittest.
This removed ugly errors with non-fitted models passed to set_model


stephanegaiffas:
@Mbompr @vinther : can you review this ?

PR #26, opened 2017-05-30
TICK-161 Added muit_incr blas implementation

vinther:


vinther:
We're running it on a macos machine. I think it has to do with the precision of the comparison inside the unit tests.

stephanegaiffas:
But, I’m on a Mac OS Machine as well...


Le 31 mai 2017 à 11:12:56, Søren Vinther Poulsen (notifications@github.com)
a écrit:

We're running it on a macos machine. I think it has to do with the
precision of the comparison inside the unit tests.

—
You are receiving this because you commented.
<https://github.com/X-DataInitiative/tick/pull/26#issuecomment-305132968>,
or mute the thread

vinther:
Yes that's my point, it works on Linux, so now we're running on macos to confirm the error :)

It's definitely some precision thing.

stephanegaiffas:
Okay then.
But is it not weird that Travis CI’s build on Mac OS is not failing in this
case ?
Error in C++ tests don’t break Travis’s builds ?
Sorry, I’m not sure to understand :)
S


Le 31 mai 2017 à 11:25:34, Søren Vinther Poulsen (notifications@github.com)
a écrit:

Yes that's my point, it works on Linux, so now we're running on macos to
confirm the error :)

It's definitely some precision thing.

—
You are receiving this because you commented.
<https://github.com/X-DataInitiative/tick/pull/26#issuecomment-305135997>,
or mute the thread

vinther:
The C++ tests did not fail on the Travis Mac OS. 

I'm thinking of two possible reasons. 

First is that the test data is randomly generated on each launch of the tests. I'm less and less sure that this was a good idea on my part.

The other is that we only test a single compiler version (whichever Travis provides), and it may produce slightly different code in terms of conversion between floats and doubles.

So yes, Travis will definitely fail if the unittests fail, but there are still some cracks where small precision errors can slip by.

stephanegaiffas:
Okay, thanks for your input :)
Maybe we should just decrease the level of tolerance in this case ?
If I recall well, the differences would only occur after 6 or 7 digits…
Cheers
S


Le 31 mai 2017 à 11:47:45, Søren Vinther Poulsen (notifications@github.com)
a écrit:

The C++ tests did not fail on the Travis Mac OS.

I'm thinking of two possible reasons.

First is that the test data is randomly generated on each launch of the
tests. I'm less and less sure that this was a good idea on my part.

The other is that we only test a single compiler version (whichever Travis
provides), and it may produce slightly different code in terms of
conversion between floats and doubles.

So yes, Travis will definitely fail if the unittests fail, but there are
still some cracks where small precision errors can slip by.

—
You are receiving this because you commented.
<https://github.com/X-DataInitiative/tick/pull/26#issuecomment-305141451>,
or mute the thread

vinther:
Yeah we already have this functionality (thanks to Martin) to test for relative errors instead of an absolute value. I've changed the unit tests to use these methods now.

PR #27, opened 2017-05-31
TICK-233 add tutorial on Hawkes on real finance (bund) data

Mbompr:


Mbompr:
This PR is necessary for tick paper

PR #28, opened 2017-06-02
TICK-374 Prox spring cleaning

stephanegaiffas:
This features :
* a is_separable method to know if a prox is separable or not
* a call on a coordinate with n_times option (useful for exact lazy updates)
* many clean, removing redundant code

### DIFF
--- a/tick/optim/prox/src/prox.cpp
+++ b/tick/optim/prox/src/prox.cpp
@@ -1,99 +1,104 @@
-//
-// Created by Martin Bompaire on 26/10/15.
-//
-
 #include "prox.h"
 
-Prox::Prox(double strength) {
+Prox::Prox(double strength,
+           bool positive) {
     has_range = false;
     this->strength = strength;
+    this->positive = positive;
 }
 
 Prox::Prox(double strength,
            ulong start,
-           ulong end) {
+           ulong end,
+           bool positive) :
+    Prox(strength, positive) {
     set_start_end(start, end);
-    this->strength = strength;
-
-    if (start > end)
-        TICK_ERROR(get_class_name() << " can't have start(" << start << ") greater than end(" << end << ")");
 }
 
 const std::string Prox::get_class_name() const {
     return "Prox";
 }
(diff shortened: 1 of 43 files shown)


tick/optim/prox/src/prox_multi.cpp
```diff
@@ -1,43 +1,38 @@
...
-
 ProxMulti::ProxMulti(std::vector<ProxPtr> proxs,
                      double strength)
-    : Prox(strength), proxs(proxs) {
-    if (strength != 0) {
-        throw std::invalid_argument("ProxMulti can be instantiated with strength=0 only, since "
-                                            "ProxMulti's strength is not used");
-    }
+  : Prox(strength, false), proxs(proxs) {
+  if (strength != 0) {
```
vinther: We shouldn't do equality comparison with floating points. If strength is a computed value it will break (but I guess it will be initialized to 0.0 all the time)

tick/optim/prox/src/prox_slope.h
```diff
@@ -1,38 +1,37 @@
...
 #include "prox.h"
 #include "prox_sorted_l1.h"
 
-
 class ProxSlope : public ProxSortedL1 {
  protected:
-    double fdr;
-
-    virtual void compute_weights(void);
+  double fdr;
```
vinther: Naming. What is fdr?

tick/optim/prox/src/prox.cpp
```diff
@@ -1,99 +1,181 @@
...
-// Created by Martin Bompaire on 26/10/15.
-//
-
 #include "prox.h"
 
-Prox::Prox(double strength) {
-    has_range = false;
-    this->strength = strength;
+Prox::Prox(double strength,
+           bool positive) {
```
vinther: Not saying it's very important here, but another time we should consider adding an enum instead of bool toggle: https://akrzemi1.wordpress.com/2017/02/16/toggles-in-functions/

PR #34, opened 2017-06-21
TICK-370 Solves a plot stems problem with bokeh

stephanegaiffas:
Currently, we latest version of bokeh, many warnings are raised

### DIFF
--- a/tick/plot/plot_stem.py
+++ b/tick/plot/plot_stem.py
@@ -1,89 +1,138 @@
 # License: BSD 3 clause
 
-import numpy as np
+
+import numpy
+from matplotlib.pyplot import Axes
 
 
-def stem(y: np.ndarray, show=True, title=None, x_range=None, y_range=None):
-    """Stem plot using bokeh
+def __stem_matplotlib(y: numpy.array, axis: Axes, title: str,
+                      x_range: tuple, y_range: tuple):
+    axis.stem(y, title=title)
+    if x_range is not None:
+        axis.set_xlim(x_range)
+    if y_range is not None:
+        axis.set_ylim(y_range)
+    if title is not None:
+        axis.set_title(title, fontsize=16)
+    return axis
 
-    Parameters
-    ----------
-    y : `np.ndarray`, shape=(n_coeffs,)
-        The vector to plot
 
-    show : `bool`, default=`True`
-        if True, show the plot. Use False if you want to superpose several
-        plots
+def __stems_matplotlib(ys: list, titles: list, x_range: tuple, y_range: tuple,
+                       fig_size):
+    from matplotlib.pyplot import figure
 
-    title : `str`, default=`None`
-        The title of the plot
+    n_stems = len(ys)
+    fig = figure(figsize=(fig_size[0], fig_size[1] * n_stems))
+    for idx, (y, title) in enumerate(zip(ys, titles)):
+        ax = fig.add_subplot(n_stems, 1, idx + 1)
+        __stem_matplotlib(y, ax, title, x_range, y_range)
+    return fig
 
-    Returns
-    -------
-    output : `bk.figure`
-        A bokeh figure object
-    """
-    import bokeh.plotting as bk
+
+def __stem_bokeh(y: numpy.array, title, x_range, y_range,
+                 fig_size):
+    import numpy as np
+    from bokeh.plotting import figure
 
     dim = y.shape[0]
     x = np.arange(dim)
-    plot_width = 600
-    plot_height = 200
-    fig = bk.figure(plot_width=plot_width, plot_height=plot_height,
-                    x_range=x_range, y_range=y_range)
+    plot_width, plot_height = fig_size
+    fig = figure(plot_width=plot_width, plot_height=plot_height,
+                 x_range=x_range, y_range=y_range)
     fig.scatter(x, y, size=4, fill_alpha=0.5)
     fig.segment(x, np.zeros(dim), x, y)
-    fig.title_text_font_size = "12pt"
+    fig.title.text_font_size = "12pt"
     if title is not None:
-        fig.title = title
-    if show:
-        bk.show(fig)
-        return None
-    else:
-        return fig
+        fig.title.text = title
+    return fig
 
 
-def stems(ys: list, titles: list = None, show=True, sync_axes=True):
-    """Several stem plots with synchronized axes
-
-    Parameters
-    ----------
-    ys : `list` of `np.ndarray`
-        A list of numpy arrays to be plotted
-
-    titles : `list` of `str`
-        The titles of each plot
-
-    show : `bool`, default=`True`
-        if True, show the plot. Use False if you want to superpose several
-        plots
-
-    sync_axes : `bool`, default=`True`
-        If True, the axes of the stem plot are synchronized
-
-    Returns
-    -------
-    output : `bk.grid_plot`
-        A grid plot object
-    """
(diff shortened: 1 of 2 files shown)


tick/plot/plot_stem.py
```diff
@@ -1,12 +1,13 @@
...
+import warnings
 
 
 def stem(y: np.ndarray, show=True, title=None, x_range=None, y_range=None):
     """Stem plot using bokeh
 
     Parameters
     ----------
-    y : `np.ndarray`, shape=(n_coeffs,)
+    y : `numpy.ndarray`, shape=(n_coeffs,)
```
Mbompr: We have been writing `np.ndarray` everywhere, should we change it?

tick/plot/plot_stem.py
```diff
@@ -31,18 +38,18 @@ def stem(y: np.ndarray, show=True, title=None, x_range=None, y_range=None):
...
     fig.segment(x, np.zeros(dim), x, y)
-    fig.title_text_font_size = "12pt"
+    fig.title.text_font_size = "12pt"
     if title is not None:
-        fig.title = title
+        fig.title.text = title
     if show:
         bk.show(fig)
-        return None
+        return fig
```
Mbompr: Should it return None of fig? With matpllotlib utilities we are returning fig everytime but it seems to be a bad idea. When you launch the function with show=True in a notebook it shows it twice...

tick/plot/plot_stem.py
```diff
@@ -1,12 +1,13 @@
 import numpy as np
+import warnings
 
 
 def stem(y: np.ndarray, show=True, title=None, x_range=None, y_range=None):
```
Mbompr: should it be called stem_bokeh?

tick/plot/plot_stem.py
```diff
@@ -85,3 +92,116 @@ def stems(ys: list, titles: list = None, show=True, sync_axes=True):
         return None
     else:
         return p
+
+
+def stems_matplotlib(ys: list, titles: list = None, show=True, sync_axes=True):
+    """Several stem plots with synchronized axes using bokeh rendering
```
Mbompr: matplotlib

tick/plot/plot_stem.py
```diff
@@ -85,3 +92,116 @@ def stems(ys: list, titles: list = None, show=True, sync_axes=True):
...
+        if titles is not None:
+            title = titles[idx]
+        else:
+            title = None
+        fig = plt.stem(y, title=title, x_range=x_range, y_range=y_range)
+        figs.append(fig)
+        if idx == 0 and sync_axes:
+            x_range = fig.x_range
+            y_range = fig.y_range
+    p = bk.gridplot([[e] for e in figs])
```
Mbompr: is it using functions from bokeh? Why?

tick/plot/plot_stem.py
```diff
@@ -85,3 +92,116 @@ def stems(ys: list, titles: list = None, show=True, sync_axes=True):
...
+    sync_axes : `bool`, default=`True`
+        If True, the axes of the stem plot are synchronized (available only
+        with ``rendering='bokeh'``
+
+    rendering : {'matplotlib', 'bokeh'}, default='matplotlib'
+        Rendering library. 'bokeh' might fail if the module is not installed.
+
+    Returns
+    -------
+    output : `bokeh.models.layouts.Column` or `None`
```
Mbompr: Output might be a matplotlib figure if rendering is set to matpllotlib isn't it?

tick/plot/plot_stem.py
```diff
@@ -85,3 +92,116 @@ def stems(ys: list, titles: list = None, show=True, sync_axes=True):
...
+    p = bk.gridplot([[e] for e in figs])
+    if show:
+        bk.show(p)
+        return None
+    else:
+        return p
+
+
+def stems(ys: list, titles: list = None, show=True, sync_axes=True,
+          rendering: str='bokeh'):
```
Mbompr: Don't we use matpllotlib by default anywhere? Bokeh is not a requirement of tick

tick/plot/plot_stem.py
```diff
@@ -85,3 +92,116 @@ def stems(ys: list, titles: list = None, show=True, sync_axes=True):
...
+        Rendering library. 'bokeh' might fail if the module is not installed.
+
+    Returns
+    -------
+    output : `bokeh.models.layouts.Column` or `None`
+        A bokeh object containing the layout of the plot, when ``show=False``,
+        None otherwise.
+    """
+
+    # TODO: put matplotlib rendering outsize. It has to return the figure if show=False
```
Mbompr: You don't use `stems_matplotlib`?

tick/plot/plot_stem.py
stephanegaiffas: If think that using np.ndarray does not link to the numpy doc in sphinx's html while numpy.ndarray does, but it needs to be confirmed :)

tick/plot/plot_stem.py
stephanegaiffas: This is why I return None here... (if I recall well)

tick/plot/plot_stem.py
stephanegaiffas: I'll correct

tick/plot/plot_stem.py
stephanegaiffas: ?

tick/plot/plot_stem.py
stephanegaiffas: Yes, default should be matplotlib, I'll change that

tick/plot/plot_stem.py
stephanegaiffas: Yes of course

tick/plot/plot_stem.py
stephanegaiffas: Don't forget this is still WIP :)

tick/plot/plot_stem.py
Mbompr: You say in the docstring you're doing so but in the code you are returning `fig`

tick/plot/plot_stem.py
Mbompr: `p = bk.gridplot([[e] for e in figs])` in `stems_matplotlib` function

tick/plot/plot_stem.py
Mbompr: Yes I've realized this afterward, but as you assigned me as a reviewer I thought it was ready

stephanegaiffas:
Ready for a review

stephanegaiffas:
Can someone reviews that ? (it's an easy and short one) @Mbompr @vinther @MaryanMorel

tick/plot/plot_stem.py
```diff
@@ -1,89 +1,138 @@
...
     Returns
     -------
-    output : `bk.grid_plot`
-        A grid plot object
+    output : `bokeh.models.layouts.Column`, `matplotlib.pyplot.Figure` or `None`
+        Depending on the rendering type, returns a ``bokeh`` or a ``matplotlib``
+        object containing the layout of the plot, or `None` otherwise
     """
-    import bokeh.plotting as bk
+    from bokeh.plotting import show as bk_show
```
Mbompr: This should be imported in the bokeh part of the code. This idea is that tick should be able to run without bokeh installed

Mbompr:
Is it used anywhere in the doc? It could be cool to ensure that the matplotlib version is always working.

PR #41, opened 2017-07-07
Tick 381 correct bug in hawkesconditiona

mrambaldi:
Now handles correctly cases where there are no events in a component.


tick/inference/hawkes_conditional_law.py
```diff
@@ -245,6 +245,10 @@ def fit(self, events: list):
             component j of realization i.
             If only one realization is given, it will be wrapped into a list
 
+        T : `double`, default=None
+            The duration (in physical time) of the realization. If it is -1 then
```
Mbompr: If it is `None`

tick/inference/hawkes_conditional_law.py
```diff
@@ -430,7 +445,7 @@ def _lexical_to_index(self, i, j, l):
         """
         return self._ijl2index[i][j][l]
 
-    def incremental_fit(self, realization, compute=True):
+    def incremental_fit(self, realization, T=None, compute=False):
```
Mbompr: I truly think `incremental_fit` should have `compute=True` by default. Default behavior of `incremental_fit` should be that model is ready after calling it. This is what has been done with other learners.

tick/inference/hawkes_conditional_law.py
```diff
@@ -476,14 +496,36 @@ def incremental_fit(self, realization, compute=True):
...
 
+        # We compute last event time
+        last_event_time = -1
+        for i in range(0, self.n_nodes):
+            if len(realization[i][0]) > 0:
+                last_event_time = max(realization[i][0][-1], last_event_time)
+
+        # If realization empty --> return
+        if last_event_time < 0:
+            return
```
Mbompr: maybe should we raise a warning here?

tick/inference/hawkes_conditional_law.py
```diff
@@ -476,14 +496,36 @@ def incremental_fit(self, realization, compute=True):
...
+
+        # We set T if needed
+        if T is None:
+            T = last_event_time
+        elif T < last_event_time:
+            raise ValueError("Argument T (%g) specified is too small (you should use default value)" % T)
+
```
Mbompr: default value or a value bigger than %g" % (T, last_event_time)

tick/inference/hawkes_conditional_law.py
```diff
@@ -367,6 +371,17 @@ def _init_lambdas(self):
         self.mean_intensity = [0] * self.n_nodes
         self._lam_N = [0] * self.n_nodes
         self._lam_T = [0] * self.n_nodes
+        # Used to store the number of events of each component that
+        # have been used to perform estimation on all the lags
+        # versus the number of events that could not be used for all the lags
+        # Warning : we don't take care of marks for this computation
+        # normally we should do this computation independantly for each mark
```
Mbompr: Is this learner still compatible with marks?

tick/inference/hawkes_conditional_law.py
```diff
@@ -516,14 +564,14 @@ def incremental_fit(self, realization, compute=True):
             self._set('_lock', pool.lock)
 
         for index, (i, j, l) in enumerate(self._index2ijl):
-            if len(realization[i][0]) == 0 or len(realization[j][0]) == 0:
-                msg = "Some realizations are empty, it is not allowed"
-                raise ValueError(msg)
+#            if len(realization[i][0]) == 0 or len(realization[j][0]) == 0:
+#                msg = "Some realizations are empty, it is not allowed"
+#                raise ValueError(msg)
```
Mbompr: is it now allowed?

tick/inference/hawkes_conditional_law.py
```diff
@@ -590,29 +638,36 @@ def incremental_fit(self, realization, compute=True):
...
-                            self.marked_components[j][l][1], claw_X, claw_Y)
+                            self.marked_components[j][l][1],
+                            T,
+                            lambda_i,
+                            claw_X, claw_Y)
 
         self._claw_X = claw_X
-        # TODO: this lock acquire is very expensive here
+
+        # TODO: this lock acquire is very expensive here  --> Do we really need it ??
```
Mbompr: Well, what happens if two threads update the claws simultaneously?

tick/inference/tests/hawkes_conditional_law_test.py
```diff
@@ -19,28 +19,29 @@ def test_hawkes_conditional_law_norm(self):
...
-                                              [-1.19782933, -1.75748751]])
+                                             [[-0.81130911, -1.12992177],
+                                              [-1.16313257, -1.72348019]])
 
     def test_hawkes_conditional_law_kernels(self):
         """...Test HawkesConditionalLaw kernel estimation
         """
         saved_phi_path = os.path.join(os.path.dirname(__file__),
                                       'hawkes_conditional_law_test-kernels.npy')
+        #np.save(saved_phi_path,self.model.kernels)
```
Mbompr: remove it or add a comment saying this should be run only if code is updated

tick/inference/hawkes_conditional_law.py
mrambaldi: yes, this was one of the point of the update. However if a component is empty on all realisations an error is raised when compute is called.

tick/inference/hawkes_conditional_law.py
mrambaldi: ok

tick/inference/hawkes_conditional_law.py
mrambaldi: I understand the need to be consistent across the library, however the whole point of `incremental_fit` with wiener-hopf is to average the conditional laws and then to compute a final estimate. With moderately high dimension the `compute` is quite slow so doing it at every update is very costly. So if `compute=True` is needed for consistency, at least we should warn the user I think...

tick/inference/hawkes_conditional_law.py
mrambaldi: yes it should be still compatible, this is just about the test used to control if a component has at least some usable events for estimation.

tick/inference/hawkes_conditional_law.py
mrambaldi: ok

tick/inference/hawkes_conditional_law.py
mrambaldi: ok

PR #44, opened 2017-07-12
TICK-384 Added alternative to view_row

vinther:


stephanegaiffas:
It this ready ?

tick/base/array/src/array2d.h
```diff
@@ -1,8 +1,6 @@
 #ifndef TICK_BASE_ARRAY_SRC_ARRAY2D_H_
 #define TICK_BASE_ARRAY_SRC_ARRAY2D_H_
 
-// License: BSD 3 clause
```
stephanegaiffas: We should leave this license line

vinther: We should replace view_row with the operator() in all places where we only use the view to access single values. Like, `view_row(...)[i]` is bad, because we construct and destroy a view for a single memory access. Also prevents compiler optimizations.

vinther: Re-added the license text :)

vinther:
We should replace view_row with the operator() in all places where we only use the view to access single values. Like, view_row(...)[i] is bad, because we construct and destroy a view for a single memory access. Also prevents compiler optimizations.

tick/base/array/src/array2d.h
```diff
@@ -112,28 +112,70 @@ class Array2d : public BaseArray2d<T> {
...
+     *
+     * @param i Row index
+     * @param j Column index
+     * @return Value at position (i, j)
+     */
+    T &operator()(const ulong i, const ulong j) {
+        const ulong idx = i * this->n_cols() + j;
+
+        #ifdef DEBUG_COSTLY_THROW
+            if (idx >= _size) TICK_BAD_INDEX(0, _size, idx);
```
Mbompr: I would rather test for i >= n_rows or j >= n_cols

PR #45, opened 2017-07-12
TICK-385 Added fast-math and O3 compilation flags

vinther:


stephanegaiffas:
I have the same error as Travis's on my laptop. 
Seems to be compiling fine, but C++ tests seem to fail

vinther:
Some silly mistake prevented the build from passing. Trying again now.

stephanegaiffas:
There is still a problem (I just pulled to get last version of the branch) :
CMake Error at CMakeLists.txt:83 (add_subdirectory):
  add_subdirectory given source "benchmark" which is not an existing
  directory.

vinther:
I think something is wrong on your end

setup.py
```diff
@@ -37,6 +37,10 @@
 
 debug_flags = ['DEBUG_COSTLY_THROW']
 
+# If true, add compilation flags to use fast (but maybe inaccurate) math
+# See https://gcc.gnu.org/wiki/FloatingPointMath
+use_fast_math = False
```
Mbompr: You've convinced me that default should be True

Mbompr:
@stephanegaiffas Everything is fine on my machine

stephanegaiffas:
Okay problem solved

stephanegaiffas:
Don't merge ! Still not happy with : some python unit tests are failing (which is perfectly normal)
I'll correct

PR #50, opened 2017-07-25
TICK-393 add hessian for hawkes exponential kernels likelihood

Mbompr:


tick/optim/model/src/hawkes_fixed_expkern_loglik.cpp
```diff
@@ -351,6 +362,44 @@ double ModelHawkesFixedExpKernLogLik::hessian_norm_dim_i(const ulong i,
...
+
+    double s = mu_i;
+    s += alpha_i.dot(g_i_k);
+    const double s_2 = s * s;
+
+    // fill mu mu
+    out[start_mu_line] += 1. / s_2;
+    // fill mu alpha
+    for (ulong j = 0; j < n_nodes; ++j) {
+      out[start_mu_line + j + 1] += g_i_k[j] / s_2;
```
vinther: There is a trick here, instead of dividing by `s_2` many times (this line and two below), we can compute` 1.0 / s_2` once and multiply it instead.

Mbompr: Oh, is this much more efficient? Isn't this things that compiler does?

tick/optim/model/src/hawkes_fixed_expkern_leastsq.h
```diff
@@ -76,8 +76,8 @@ class ModelHawkesFixedExpKernLeastSq : public ModelHawkesSingle {
   void grad_i(const ulong i, const ArrayDouble &coeffs, ArrayDouble &out) override;
```
Mbompr: Hum, I think it computes grad_i, documentation is for the function that comes after void hessian(ArrayDouble &out)

tick/optim/model/base/model_hawkes.py
```diff
@@ -1,7 +1,10 @@
 # License: BSD 3 clause
 
 import numpy as np
+from scipy.sparse import sputils, csr_matrix
 
+from tick.optim.model.build.model import (ModelHawkesFixedSumExpKernLeastSqList,
```
Mbompr: I find it cleaner than many `\`, I've seen it in another open source repo, probably scikit-learn

tick/optim/model/src/hawkes_fixed_expkern_loglik.cpp
Mbompr: We've conclude that --fast-math does this so it can be left :)

tick/optim/model/base/model_hawkes.py
stephanegaiffas: OK !

tick/optim/model/src/hawkes_fixed_expkern_leastsq.h
stephanegaiffas: Oh okay !

PR #56, opened 2017-08-04
TICK-397 Building on windows

Dekken:
There's assumptions that MSVC is in use. 

Installing the requirements from pip is not so straight forward either and requires downloading some WHL files separately

```
numpy-1.13.1+mkl-cp36-cp36m-win_amd64.whl
numpydoc-0.7.0-py2.py3-none-any.whl
scipy-0.19.1-cp36-cp36m-win_amd64.whl
```

Which I got from here: www.lfd.uci.edu/~gohlke/pythonlibs

There's an issue linking libraries so it has to be done manually. 

I haven't tested linux and mac still work but I don't see why they wouldn't.

Feel free to reject and do manually.

Most notable change is probably in:
```
tick/base/src/interruption.h
```

As sharing static class members across DLLs is tricky, it's changed to a static method/member pair.

Prebuilt windows libraries can be found on the windows branch of my fork under the "build" directory.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -285,13 +285,16 @@ def create_extension(extension_name, module_dir,
     else:
         extra_compile_args.extend(min_extra_compile_args)
 
-    # Added -Wall to get all warnings and -Werror to treat them as errors
-    extra_compile_args.append("-Wall")
-    extra_compile_args.append("-Werror")
+    if os.name == 'nt':
+        extra_compile_args.append("-DBUILDING_DLL")
+    else:
+        # Added -Wall to get all warnings and -Werror to treat them as errors
+        extra_compile_args.append("-Wall")
+        extra_compile_args.append("-Werror")
 
-    # This warning is turned off because SWIG generates files that triggers the
-    # warning
-    extra_compile_args.append("-Wno-uninitialized")
+        # This warning is turned off because SWIG generates files that triggers the
+        # warning
+        extra_compile_args.append("-Wno-uninitialized")
 
     # Include directory of module
     mod = SwigPath(module_dir, extension_name)
@@ -339,6 +342,8 @@ def create_extension(extension_name, module_dir,
             # $ORIGIN refers to the location of the current shared object file
             # at runtime
             runtime_library_dirs.append("\$ORIGIN/%s" % rel_path)
+        elif os.name == 'nt':
+            pass
         else:  # Assuming non-Windows builds for now
             # For OSX builds we use @loader_path instead
             extra_link_args.append(
@@ -351,6 +356,8 @@ def create_extension(extension_name, module_dir,
     filename = swig_path.lib_filename
     if platform.system() == 'Linux':
         extra_link_args.append('-Wl,-soname,%s' % filename)
+    elif os.name == 'nt':
+        pass
     else:
         # For OSX the install_name needs to be prefixed with @rpath
         extra_link_args.append('-Wl,-install_name,@rpath/%s' % filename)
--- a/tick/base/src/interruption.h
+++ b/tick/base/src/interruption.h
@@ -11,6 +11,8 @@
 
 // License: BSD 3 clause
 
+#include "defs.h"
+
 #include <atomic>
 #include <exception>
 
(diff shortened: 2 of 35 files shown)


Mbompr:
Well, that's pretty clean!! Travis will tell us what he thinks about mac and linux builds. I'd love to merge that!

vinther:
Travis builds fail over cpplint checks

Mbompr:
Very minor changes are required to fit with google style
```
tick/base/src/interruption.h:24:  Missing space before {  [whitespace/braces] [5]
tick/base/src/interruption.h:39:  Line ends in whitespace.  Consider deleting these extra spaces.  [whitespace/end_of_line] [4]
tick/base/src/interruption.h:46:  Line ends in whitespace.  Consider deleting these extra spaces.  [whitespace/end_of_line] [4]
tick/base/src/defs.h:51:  Use int16/int64/etc, rather than the C type short  [runtime/int] [4]
tick/base/src/defs.h:56:  At least two spaces is best between code and comments  [whitespace/comments] [2]
tick/base/src/defs.h:56:  Should have a space between // and comment  [whitespace/comments] [4]
tick/base/src/debug.h:29:  At least two spaces is best between code and comments  [whitespace/comments] [2]
tick/base/src/debug.h:29:  Should have a space between // and comment  [whitespace/comments] [4]
tick/base/src/debug.h:31:  At least two spaces is best between code and comments  [whitespace/comments] [2]
tick/base/src/debug.h:31:  Should have a space between // and comment  [whitespace/comments] [4]
tick/base/src/debug.h:64:  Redundant blank line at the start of a code block should be deleted.  [whitespace/blank_line] [2]
tick/base/src/debug.h:65:  Line ends in whitespace.  Consider deleting these extra spaces.  [whitespace/end_of_line] [4]
```

Dekken:
Thanks Martin, I'll make those changes later today.

Dekken:
I had an ifdef "HAVE_EXECINFO" in there but it should have been "HAVE_EXECINFO_H", but it's only relevant if you're support NetBSD (libexecinfo has to be linked manually), so I removed it

setup.py
```diff
@@ -285,13 +285,16 @@ def add_dir_name(dir_name, filenames):
     else:
         extra_compile_args.extend(min_extra_compile_args)
 
-    # Added -Wall to get all warnings and -Werror to treat them as errors
-    extra_compile_args.append("-Wall")
-    extra_compile_args.append("-Werror")
+    if os.name == 'nt':
+        extra_compile_args.append("-DBUILDING_DLL")
```
stephanegaiffas: No -Wall -Werror for the nt compilation ?

setup.py
```diff
@@ -351,6 +356,8 @@ def add_dir_name(dir_name, filenames):
     filename = swig_path.lib_filename
     if platform.system() == 'Linux':
```
stephanegaiffas: Seems that we use both os.name and platform.system() to determine the OS.
Which one should be actually used ? It's probably best to use a single one to have consistent results ?

setup.py
Dekken: Werror doesn't exist for MSVC, and the command already includes "/W3" by default from distutils. but I can put "-Wall" back in sure.

tick/base/src/interruption.h
```diff
@@ -19,31 +21,40 @@
...
 
-    const char *what() const noexcept override;
+    DLL_PUBLIC const char *what() const noexcept override;
 
     //! \cond
     //! @brief Notifies an interruption has been detected (not to be called directly)
-    inline static void set() { flag_interrupt = true; }
+    inline static void set() {
+        auto& flag_interrupt = get_flag_interrupt();
+        flag_interrupt = true;
```
stephanegaiffas: Why are we getting the flag_interrupt, and overwriting it right after ? (there must be a mechanism that I'm not aware of :))

setup.py
Dekken: platform.system() is better, as it's "windows" rather than just "nt" for os.name which isn't very clear

setup.py
stephanegaiffas: Okay :)

setup.py
stephanegaiffas: Okay, so let's stick to a single way of determining the OS then, I think that it's good to have a single consistent way

tick/base/src/interruption.h
Dekken: the static member was replaced by a static function containing a static member as static members across DLLs don't work so well, it could be fine to just have 


````
get_flag_interrupt() = true;
```

stephanegaiffas: Oh okay then. It's fine the way it is !

stephanegaiffas:
This really is an important PR. If this allows indeed to compile on windows (I cannot test it, but I trust Philip), I think that this PR should also contain an update of the README.md to explain that in order to use tick on windows, only the git cloning and compilation works for now, and that some dependencies have to installed by hand (if I understand well). The README should provide some instructions so that a motivated user can follow some steps to get it compiling and working on windows.

After this is merged, we should spend some time on getting travis testing for windows (for the three python version we test upon) and, later, manage to produce binaries as we do with macOS and linux (I know that this can be painful, so let's plan that for later :))

I'll add a JIRA task and update the naming of the PR to make it consistent with our workflow.

Once again, that really pretty cool !

tick/optim/model/src/coxreg_partial_lik.h
```diff
@@ -9,8 +9,7 @@
 
 #include "model.h"
 
-
-class ModelCoxRegPartialLik : public Model {
+class DLL_PUBLIC ModelCoxRegPartialLik : public Model {
```
Mbompr: If I understand correctly, this flag must be put if the class is shared with another module. It this class shared out of model module?

Mbompr: This is the same for all model classes except "Hawkes" ones. I've seen that for Prox you've only done it with ProxZero and its ancestors.

Dekken: the inference module file "inference_module_wrap.cpp" uses a number of model classes

Mbompr: Oh that's true, I have no idea why they appear there but that's another problem!

Dekken:
@stephanegaiffas 
There's one big problem with building on windows, which causes linking to fail when there's any library dependencies as I mentioned earlier this results in many failed calls to "python setup.py install" and each library must be linked manually after fixing the command line call to "link.exe"

The file type which is linked during the build is ".pyd", but it should be ".lib" 

I will document the setup steps for windows with MSVC, but this is a bit of a blocker for anyone but the most ardent users.

stephanegaiffas:
Hey @Dekken I'm not familiar enough with compilation on windows... so I'm not sure to understand the problem. However, we are trusting you on this :)
I think that you can just explain the setup steps in the README, even if it's intended for motivated users :)

Dekken:
Closing and recreating from different branch name

PR #61, opened 2017-08-09
TICK-373 Full sparse support in SVRG

stephanegaiffas:
Strong improvement of SVRG for sparse data, this adds :
* Fully sparse updates in the case of sparse data, when the prox is separable. 

For now, SVRG only supports the so-called 'probabilistic' approach, where a correction of the step-sizes is used for each feature.
The exact updates, based on delayed computations, will be added in a future version.

### DIFF
--- a/tick/optim/model/src/model_labels_features.cpp
+++ b/tick/optim/model/src/model_labels_features.cpp
@@ -1,9 +1,5 @@
 // License: BSD 3 clause
 
-//
-// Created by Stéphane GAIFFAS on 06/12/2015.
-//
-
 #include "model_labels_features.h"
 
 ModelLabelsFeatures::ModelLabelsFeatures(SBaseArrayDouble2dPtr features,
@@ -11,11 +7,32 @@ ModelLabelsFeatures::ModelLabelsFeatures(SBaseArrayDouble2dPtr features,
     : n_samples(labels.get() ? labels->size() : 0),
       n_features(features.get() ? features->n_cols() : 0),
       labels(labels),
-      features(features) {
+      features(features),
+      ready_columns_sparsity(false) {
   if (labels.get() && labels->size() != features->n_rows()) {
     std::stringstream ss;
     ss << "In ModelLabelsFeatures, number of labels is " << labels->size();
     ss << " while the features matrix has " << features->n_rows() << " rows.";
     throw std::invalid_argument(ss.str());
+  }
+}
+
+void ModelLabelsFeatures::compute_columns_sparsity() {
+  if (features->is_sparse()) {
+    column_sparsity = ArrayDouble(n_features);
+    column_sparsity.fill(0.);
+    for (ulong i = 0; i < n_samples; ++i) {
+      BaseArrayDouble features_i = get_features(i);
+      for (ulong j = 0; j < features_i.size_sparse(); ++j) {
+        // Even if the entry is zero (nothing forbids to store zeros...) increment
+        // the number of non-zeros of the columns. This is necessary when computed step-size corrections
+        // used in probabilistic updates (see SVRG::solve_sparse_proba_updates code for instance)
+        column_sparsity[features_i.indices()[j]] += 1;
+      }
+    }
+    column_sparsity.multiply(1. / n_samples);
+    ready_columns_sparsity = true;
+  } else {
+    TICK_ERROR("The features matrix is not sparse.")
   }
 }
--- a/tick/optim/model/src/model_labels_features.h
+++ b/tick/optim/model/src/model_labels_features.h
@@ -1,7 +1,3 @@
-//
-// Created by Stéphane GAIFFAS on 06/12/2015.
-//
-
 #ifndef TICK_OPTIM_MODEL_SRC_MODEL_LABELS_FEATURES_H_
 #define TICK_OPTIM_MODEL_SRC_MODEL_LABELS_FEATURES_H_
 
@@ -20,6 +16,9 @@ class ModelLabelsFeatures : public virtual Model {
 
   //! Features matrix (either sparse or not)
   SBaseArrayDouble2dPtr features;
+
+  bool ready_columns_sparsity;
+  ArrayDouble column_sparsity;
 
  public:
   ModelLabelsFeatures(SBaseArrayDouble2dPtr features,
@@ -54,9 +53,20 @@ class ModelLabelsFeatures : public virtual Model {
     return n_samples;
   }
 
+  bool is_ready_columns_sparsity() const {
+    return ready_columns_sparsity;
+  }
+
+  ArrayDouble get_column_sparsity_view() {
+    if (!is_ready_columns_sparsity()) compute_columns_sparsity();
+    return view(column_sparsity);
+  }
+
+  void compute_columns_sparsity();
+
   template<class Archive>
-  void load(Archive & ar) {
-    ar(CEREAL_NVP(n_samples) );
+  void load(Archive &ar) {
+    ar(CEREAL_NVP(n_samples));
 
     ArrayDouble temp_labels;
     ArrayDouble2d temp_features;
@@ -68,7 +78,7 @@ class ModelLabelsFeatures : public virtual Model {
   }
 
   template<class Archive>
-  void save(Archive & ar) const {
+  void save(Archive &ar) const {
     ar(CEREAL_NVP(n_samples));
     ar(cereal::make_nvp("labels", *labels));
     ar(cereal::make_nvp("features", *features));
--- a/tick/optim/solver/src/svrg.cpp
+++ b/tick/optim/solver/src/svrg.cpp
@@ -1,9 +1,6 @@
 // License: BSD 3 clause
 
-//
-// Created by Martin Bompaire on 23/10/15.
-//
-
+#include "model_labels_features.h"
 #include "svrg.h"
 
 SVRG::SVRG(ulong epoch_size,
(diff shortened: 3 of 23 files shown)


stephanegaiffas:
@Mbompr : this is ready :) This one has been challenging :)

tick/optim/solver/src/svrg.cpp
```diff
@@ -1,127 +1,257 @@
...
-    if (variance_reduction == VarianceReductionMethod::Random) {
-        rand_index = rand_unif(epoch_size);
+void SVRG::solve_sparse_proba_updates() {
+  // Data is sparse, and we use the probabilistic update strategy
+  // This means that the model is a child of ModelGeneralizedLinear.
+  // The strategy used here uses non-delayed updates, with corrected
+  // step-sizes using a probabilistic approximation and the
+  // penalization trick: with such a model and prox (if prox is separable),
+  // we can work only inside the current support (non-zero values) of
+  // the sampled vector of features
```
Mbompr: Maybe can we add a link to ASAGA paper or any other paper that introduced this update type?

tick/optim/solver/src/svrg.cpp
```diff
@@ -1,127 +1,257 @@
...
+      }
+      // Update last_time
+      last_time[j] = t;
     }
-
-    if (variance_reduction == VarianceReductionMethod::Last)
-        next_iterate = iterate;
+    if (!is_prox_separable) {
+      // If prox is not separable, apply it in a non-delayed fashion
+      prox->call(iterate, step, iterate);
```
Mbompr: Is this correct?

I think this will apply a prox on all coeffs of iterate but only coeffs that are in the support of x_i are would be up to date.

tick/optim/solver/svrg.py
```diff
@@ -51,14 +57,25 @@ class SVRG(SolverFirstOrderSto):
         Information along iteration is recorded in history each time the
         iteration number of a multiple of ``record_every``
```
Mbompr: Maybe `variance_reduction` and `delayed_updates` are technical enough to go in an Other Parameters section?

tick/optim/solver/svrg.py
```diff
@@ -90,20 +108,64 @@ def __init__(self, step: float = None, epoch_size: int = None,
...
-                'variance_reduction should be one of "{}", got "{}".'.format(
-                    ', '.join(variance_reduction_methods_mapper.keys()),
+                'variance_reduction should be one of "{}", got "{}"'.format(
+                    ', '.join(sorted(variance_reduction_methods_mapper.keys())),
                     val))
-
+        if self.model is not None:
+            if val == 'avg' and self.model._model.is_sparse():
+                raise ValueError("'avg' variance reduction cannot be used "
+                                 "with delayed updates for sparse data")
```
Mbompr: I agree it would be a terrible choice but shouldn't we rather raise a warning?

tick/optim/solver/tests/svrg_test.py
```diff
@@ -54,8 +85,105 @@ def test_variance_reduction_setting(self):
         self.assertEqual(svrg._solver.get_variance_reduction(),
                          _SVRG.VarianceReductionMethod_Last)
 
-        with self.assertRaises(ValueError):
-            svrg.variance_reduction = 'wrong_name'
+        msg = '^variance_reduction should be one of "avg, last, rand", ' \
+              'got "truc"$'
```
Mbompr: Oh no, now everybody will know that we're French :)

tick/optim/solver/tests/svrg_test.py
```diff
@@ -54,8 +85,105 @@ def test_variance_reduction_setting(self):
...
+        rand_types = ['unif', 'perm']
+        delayed_updates = ['exact', 'proba']
+        proxs = [prox_l1, prox_tv]
+        solutions = []
+
+        products = product(variance_reductions, rand_types, delayed_updates,
+                           proxs)
+        for variance_reduction, rand_type, delayed_update, prox in products:
+            solver = SVRG(step=step, tol=1e-10, max_iter=30, verbose=False,
+                          print_every=1,
```
Mbompr: print_every is not necessary

tick/optim/model/src/model_labels_features.cpp
```diff
@@ -11,11 +11,36 @@ ModelLabelsFeatures::ModelLabelsFeatures(SBaseArrayDouble2dPtr features,
...
   if (labels.get() && labels->size() != features->n_rows()) {
     std::stringstream ss;
     ss << "In ModelLabelsFeatures, number of labels is " << labels->size();
     ss << " while the features matrix has " << features->n_rows() << " rows.";
     throw std::invalid_argument(ss.str());
   }
 }
+
+
+void ModelLabelsFeatures::compute_columns_non_zeros(ArrayULong &columns_sparsity) {
```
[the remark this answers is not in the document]
stephanegaiffas: Done

tick/optim/model/src/model_labels_features.cpp
```diff
@@ -11,11 +11,36 @@ ModelLabelsFeatures::ModelLabelsFeatures(SBaseArrayDouble2dPtr features,
...
+  if (!ready_columns_sparsity) {
```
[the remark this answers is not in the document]
stephanegaiffas: Done

tick/optim/model/src/model_labels_features.cpp
```diff
@@ -11,11 +11,36 @@ ModelLabelsFeatures::ModelLabelsFeatures(SBaseArrayDouble2dPtr features,
...
+      if (columns_sparsity.size() != n_features) {
+        TICK_ERROR("given `columns_sparsity` vector must match `n_features`")
+      }
+      columns_sparsity.fill(0.);
+      for (ulong i = 0; i < n_samples; ++i) {
+        BaseArrayDouble features_i = get_features(i);
+        for (ulong j = 0; j < features_i.size_sparse(); ++j) {
+          // If the entry is indeed non-zero (nothing forbids to store zeros...) increment
+          // the number of non-zeros of the columns
+          if (std::abs(features_i.data()[j]) > 0) {
```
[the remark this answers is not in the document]
stephanegaiffas: Indeed

tick/optim/model/src/model_labels_features.cpp
```diff
@@ -11,11 +11,36 @@ ModelLabelsFeatures::ModelLabelsFeatures(SBaseArrayDouble2dPtr features,
...
+            columns_sparsity[features_i.indices()[j]] += 1;
+          }
+        }
+      }
+    } else {
+      TICK_ERROR("The features matrix is not sparse.")
+    }
+  }
```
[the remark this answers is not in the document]
stephanegaiffas: Done

tick/optim/solver/src/svrg.cpp
stephanegaiffas: This is part of the task about solvers documentation (that I started already)

tick/optim/solver/src/svrg.h
```diff
@@ -10,49 +6,100 @@
...
+  ArrayDouble grad_i_fixed_w;
+  ArrayDouble next_iterate;
+
+  ulong rand_index;
+  ulong n_features;
+
+  bool ready_step_corrections;
+  bool use_intercept;
+  bool is_model_sparse;
+  bool is_prox_separable;
```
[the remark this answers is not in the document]
stephanegaiffas: I've removed most of these attributes, and pass them now to the specialized solve methods. However the ones really used in the class are kept.

tick/optim/solver/src/svrg.cpp
```diff
@@ -1,127 +1,257 @@
...
-    ulong n_features = model->get_n_features();
-    bool use_intercept = model->use_intercept();
-
-    ArrayDouble mu(iterate.size());
-    ArrayDouble fixed_w = iterate;
-    model->grad(fixed_w, mu);
-
-    ulong rand_index{0};
+void SVRG::compute_step_corrections() {
+  if (!ready_step_corrections) {
```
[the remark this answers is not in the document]
stephanegaiffas: I corrected that

tick/optim/solver/svrg.py
stephanegaiffas: Yes indeed, I've moved them

tick/optim/solver/svrg.py
stephanegaiffas: It's raising only a warning now (and unittested)

tick/optim/solver/tests/svrg_test.py
stephanegaiffas: Replaced truc by stuff

tick/optim/solver/tests/svrg_test.py
stephanegaiffas: OK

stephanegaiffas:
@Mbompr Ready for second round

tick/optim/solver/src/svrg.cpp
Mbompr: Alright!

svrg_lazy_updates.py
```diff
@@ -0,0 +1,211 @@
+
```
Mbompr: This file should probably be removed :)

stephanegaiffas:
@Mbompr : ready for third round !

tick/optim/model/src/model_labels_features.cpp
```diff
@@ -1,21 +1,39 @@
-// License: BSD 3 clause
-
-//
-// Created by Stéphane GAIFFAS on 06/12/2015.
-//
-
 #include "model_labels_features.h"
 
+// License: BSD 3 clause
```
Mbompr: It doesn't worth pushing a new commit for this, but usually, in cpp files, licence is a the very beginning of the file

tick/optim/model/src/model_labels_features.cpp
```diff
@@ -1,21 +1,39 @@
...
+      TICK_ERROR("given `out_columns_non_zeros` vector must match `n_features`")
+    }
+    out_columns_non_zeros.fill(0.);
+    for (ulong i = 0; i < n_samples; ++i) {
+      BaseArrayDouble features_i = get_features(i);
+      for (ulong j = 0; j < features_i.size_sparse(); ++j) {
+        // If the entry is indeed non-zero (nothing forbids to store zeros...) increment
+        // the number of non-zeros of the columns
+        if (features_i.data()[j] != 0) {
+          out_columns_non_zeros[features_i.indices()[j]] += 1;
```
Mbompr: Just a quick thought,
If an entry in a sparse matrix is 0 and we discard it here we should take it into account in SVRG code as well... Probably is it not such a good idea to discard it here...

Mbompr: Indeed it leads to false results

```python
    def test_dense_and_sparse_match_fake_non_zeros(self):
        """...Test in SVRG that dense and sparse code matches in all possible
        settings
        """
        variance_reductions = ['last', 'rand']
        rand_types = ['unif', 'perm']
        seed = 123
        tol = 0.
        max_iter = 20
        X, y = self.simu_linreg_data(interc=-1.)
        d = X.shape[1]

        proxs = [
            ProxTV(strength=1e-2, range=(0, d)),
            ProxElasticNet(strength=1e-2, ratio=0.9, range=(0, d)),
            ProxEquality(range=(0, d))
        ]

        # Let's create model manually
        X_sparse = csr_matrix(X)
        # fill half of the non zeros with zeros
        n_new_zeros = int(len(X_sparse.data) / 2)
        mask = np.random.choice(len(X_sparse.data), n_new_zeros, replace=False)
        # next line makes SVRG fail...
        X_sparse.data[mask] = 0
        X_dense = X_sparse.toarray()
        model_dense = ModelLinReg().fit(X_dense, y)
        model_spars = ModelLinReg().fit(X_sparse, y)

        step = 1 / model_spars.get_lip_max()
        for variance_reduction, rand_type, prox in product(variance_reductions,
                                                           rand_types, proxs):
            solver_sparse = SVRG(step=step, tol=tol, max_iter=max_iter,
                                 verbose=False,
                                 variance_reduction=variance_reduction,
                                 rand_type=rand_type, seed=seed) \
                .set_model(model_spars) \
                .set_prox(prox)

            solver_dense = SVRG(step=step, tol=tol, max_iter=max_iter,
                                verbose=False,
                                variance_reduction=variance_reduction,
                                rand_type=rand_type, seed=seed) \
                .set_model(model_dense) \
                .set_prox(prox)

            solver_sparse.solve()
            solver_dense.solve()
            np.testing.assert_array_almost_equal(solver_sparse.solution,
                                                 solver_dense.solution, 7)
```

tick/optim/solver/src/svrg.cpp
```diff
@@ -1,127 +1,150 @@
 // License: BSD 3 clause
 
-//
-// Created by Martin Bompaire on 23/10/15.
-//
-
+#include <model_labels_features.h>
```
Mbompr: include with "" instead of <> (XCode compatibility)

tick/optim/model/src/model_labels_features.h
```diff
@@ -21,6 +17,10 @@ class ModelLabelsFeatures : public virtual Model {
   //! Features matrix (either sparse or not)
   SBaseArrayDouble2dPtr features;
 
+  bool ready_columns_sparsity;
+
+  ArrayULong columns_sparsity;
```
Mbompr: This array is not used anywhere is it?

tick/optim/solver/src/svrg.cpp
```diff
@@ -1,127 +1,150 @@
...
+  // Data is sparse, and we use the probabilistic update strategy
+  // This means that the model is a child of ModelGeneralizedLinear.
+  // The strategy used here uses non-delayed updates, with corrected
+  // step-sizes using a probabilistic approximation and the
+  // penalization trick: with such a model and prox, we can work only inside the current
+  // support (non-zero values) of the sampled vector of features
+  std::shared_ptr<ProxSeparable> casted_prox;
+  if (prox->is_separable()) {
+    casted_prox = std::static_pointer_cast<ProxSeparable>(prox);
+  }
```
Mbompr: else raise an error

Otherwise we will have a segfault (sorry I should have added it)

tick/optim/solver/src/svrg.cpp
```diff
@@ -1,127 +1,150 @@
...
+      // Step-size correction for coordinate j
+      double step_correction = steps_correction[j];
+      // Gradient descent with probabilistic step-size correction
+      iterate[j] -= step * (x_i.data()[idx_nnz] * grad_i_diff + step_correction * full_gradient_j);
+      // Prox is separable, apply regularization on the current coordinate
+      iterate[j] = casted_prox->call_single(iterate[j], step * step_correction);
+    }
+    // And let's not forget to update the intercept as well
+    // It's updated at each step, so no step-correction and no prox applied it
+    if (use_intercept) {
```
Mbompr: Why don't we apply call_single to it? If it's out of the range it should not be modified and if it is in (even if this is not the best idea ever) it would be consistent with the dense case.

tick/optim/solver/src/svrg.cpp
```diff
@@ -1,127 +1,150 @@
...
+      iterate[n_features] -= step * (grad_i_diff + full_gradient[n_features]);
+    }
+    // Note that the average option for variance reduction with sparse data is a very bad idea,
+    // but this is catched in the python class
```
Mbompr: caught

tick/optim/solver/svrg.py
```diff
@@ -51,12 +53,15 @@ class SVRG(SolverFirstOrderSto):
...
+    Other Parameters
+    ----------------
+    variance_reduction : {'last', 'avg', 'rand'}, default='last'
         Determine what is used as phase iterate for variance reduction.
 
         * 'last' : the phase iterate is the last iterate of the previous epoch
         * 'avg' : the phase iterate is the average over the iterates in the past
-          epoch
+          epoch. This is a really bad idea when using delayed updates (with
+          sparse datasets, a warning will be raised in this case)
```
Mbompr: This is a really bad idea when using sparse datasets, a warning will be raised in this case

Mbompr: We don't mention delayed updates anymore

tick/optim/solver/svrg.py
```diff
@@ -98,12 +103,38 @@ def variance_reduction(self):
...
+            order information about the model (loss, gradient, among
+            other things)
+
+        Returns
+        -------
+        output : `Solver`
+            The `Solver` with given model
+        """
+        # We need to check that the setted model is not sparse, while the
+        # variance reduction method is 'avg
```
Mbompr: missing '

tick/optim/solver/svrg.py
```diff
@@ -98,12 +103,38 @@ def variance_reduction(self):
...
+        if self.variance_reduction == 'avg' and model._model.is_sparse():
+            warn("'avg' variance reduction cannot be used with delayed updates "
```
Mbompr: again, no need to mention delayed updates

tick/optim/solver/tests/src/test_svrg.cpp
```diff
@@ -0,0 +1,115 @@
...
+  for (int j = 0; j < 30; ++j) {
+    svrg.solve();
+  }
+  svrg.get_iterate(out_iterate60);
+
+  out_iterate60.mult_incr(out_iterate30, -1);
+  EXPECT_LE(out_iterate60.norm_sq() / n_features, 0.1);
+}
+
+TEST(SVRG, test_sparse_exact_updates) {
```
Mbompr: This last test should be deleted

tick/optim/solver/tests/svrg_test.py
```diff
@@ -54,8 +88,119 @@ def test_variance_reduction_setting(self):
...
+                                verbose=False,
+                                variance_reduction=variance_reduction,
+                                rand_type=rand_type, seed=seed) \
+                .set_model(model_dense) \
+                .set_prox(prox)
+
+            solver_sparse.solve()
+            solver_dense.solve()
+            np.testing.assert_array_almost_equal(solver_sparse.solution,
+                                                 solver_dense.solution, 7)
```
Mbompr: We should be able to duplicate less code in this example (loop on the intercept?)

tick/optim/model/src/model_labels_features.cpp
stephanegaiffas: Nice catch ! I'll remove the test then

tick/optim/solver/src/svrg.cpp
stephanegaiffas: Done

tick/optim/solver/src/svrg.cpp
stephanegaiffas: Done

tick/optim/solver/src/svrg.cpp
stephanegaiffas: I'm not sure to understand. We can only use the call_single, that applies on a single coordinate, and this call_single does not use the range of the prox... Since we don't want to regularize the intercept, we just don't call call_single in this step.
Note that dense and sparse cases must match both with and without an intercept in the unit tests.

tick/optim/solver/src/svrg.cpp
stephanegaiffas: Corrected

tick/optim/model/src/model_labels_features.h
stephanegaiffas: Nice catch !

tick/optim/solver/src/svrg.cpp
Mbompr: Oh ok, what would happen if someone sets a prox to a defined range then? Won't we have different results in the dense and the sparse case?
Maybe call_single should take a (maybe optional) argument index that will determine if prox should be applied ot not?

tick/optim/solver/svrg.py
stephanegaiffas: Done

tick/optim/solver/svrg.py
stephanegaiffas: Done

tick/optim/solver/svrg.py
stephanegaiffas: Done

tick/optim/solver/tests/svrg_test.py
stephanegaiffas: Do we really care about code duplication in unit tests ?

tick/optim/solver/tests/src/test_svrg.cpp
stephanegaiffas: I removed it

tick/optim/solver/src/svrg.cpp
stephanegaiffas: Ouch shit you're right, once again !

tick/optim/solver/tests/svrg_test.py
```diff
@@ -54,8 +88,119 @@ def test_variance_reduction_setting(self):
...
+        """
+        variance_reductions = ['last', 'rand']
+        rand_types = ['unif', 'perm']
+        seed = 123
+        tol = 0.
+        max_iter = 20
+        X, y = self.simu_linreg_data(interc=-1.)
+        d = X.shape[1]
+
+        proxs = [
```
Mbompr: Add exotic ranges to prox

tick/optim/solver/tests/svrg_test.py
```diff
@@ -54,8 +88,119 @@ def test_variance_reduction_setting(self):
...
+            ProxTV(strength=1e-2, range=(0, d)),
+            ProxElasticNet(strength=1e-2, ratio=0.9, range=(0, d)),
+            ProxEquality(range=(0, d))
+        ]
+
```
Mbompr: add comments with and without intercepts

PR #69, opened 2017-09-14
TICK-412 Corrects things with ModelLinRegWithOutliers and RobustLinearRegression 

stephanegaiffas:
Corrects computations in ModelLinRegWithOutliers (global intercept is correctly positioned now), adds many unittests and corrected RobustLinearRegression as well (and added unittests). Adds also a unittest for GLM models that checks computations with intercept and without with hard-coded intercept

### DIFF
--- a/tick/inference/tests/robust_linear_regression_test.py
+++ b/tick/inference/tests/robust_linear_regression_test.py
@@ -1,9 +1,7 @@
 # License: BSD 3 clause
 
 import unittest
-
 import itertools
-from itertools import product
 import numpy as np
 
 from tick.inference.tests.inference import InferenceTest
@@ -13,10 +11,10 @@ from tick.metrics import support_fdp, support_recall
 
 
 class Test(InferenceTest):
-    n_samples = 1000
+    n_samples = 300
     n_features = 5
     noise_level = 1.
-    nnz_outliers = 50
+    nnz_outliers = 10
     outliers_intensity = 5.
     interc0 = -3.
     target_fdr = 0.2
@@ -30,7 +28,7 @@ class Test(InferenceTest):
 
     @staticmethod
     def get_train_data(fit_intercept=True):
-        np.random.seed(1)
+        np.random.seed(12)
         n_samples = Test.n_samples
         n_features = Test.n_features
         noise_level = Test.noise_level
@@ -51,7 +49,6 @@ class Test(InferenceTest):
               * np.sign(sample_intercepts0[sample_intercepts0 != 0])
 
         X = features_normal_cov_toeplitz(n_samples, n_features, 0.5)
-        X /= np.linalg.norm(X, axis=0)
         y = X.dot(weights0) + noise_level * np.random.randn(n_samples) + interc0
         y += sample_intercepts0
         return X, y, weights0, interc0, sample_intercepts0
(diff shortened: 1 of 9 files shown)


stephanegaiffas:
@Mbompr : can you take between 0 and 2 minutes to review this ?

tick/inference/tests/robust_linear_regression_test.py
```diff
@@ -60,53 +57,48 @@ def test_RobustLinearRegression_fit(self):
...
-        for i, (solver, penalty) \
-                in enumerate(product(RobustLinearRegression._solvers.keys(),
-                                     RobustLinearRegression._penalties.keys())):
+        for i, solver in enumerate(solvers):
             learner_keywords = {
                 'C_sample_intercepts': Test.n_samples / Test.noise_level,
                 'fit_intercept': False, 'fdr': Test.target_fdr,
-                'max_iter': 100, 'tol': 1e-10, 'solver': solver,
-                'penalty': penalty, 'verbose': False
+                'max_iter': 3000, 'tol': 1e-7, 'solver': solver,
```
Mbompr: How long takes this test to run? I think test should be < 300 ms if possible and 2sec in the worst case.

tick/optim/model/tests/generalized_linear_model.py
```diff
@@ -51,3 +51,23 @@ def run_test_for_glm(self, model, model_spars=None,
...
+                X_with_ones = np.hstack((X, np.ones((model.n_samples, 1))))
+                model.fit_intercept = False
+                model.fit(X_with_ones, y)
+                grad2 = model.grad(coeffs)
+
+                np.testing.assert_almost_equal(grad1, grad2, decimal=10)
+
+                # Put back model to its previous status
+                model.fit_intercept = True
+                model.fit(X, y)
```
Mbompr: Can we isolate this test? The merged thing is a verrry big test. We are not true unit testers but still, we try to pretend we are :)

tick/optim/model/tests/linreg_with_intercepts_test.py
```diff
@@ -13,27 +13,82 @@
...
+            .fit(X_spars, y)
         self.run_test_for_glm(model, model_spars, 1e-4, 1e-4)
 
         self.assertAlmostEqual(model.get_lip_mean(), 6.324960325598532)
         self.assertAlmostEqual(model.get_lip_max(), 30.277118951892113)
         self.assertAlmostEqual(model.get_lip_mean(), model_spars.get_lip_mean())
         self.assertAlmostEqual(model.get_lip_max(), model_spars.get_lip_max())
+        self.assertAlmostEqual(model.get_lip_best(), 2.7217793249045439)
+
+        # *** Tests with global intercept ***
```
Mbompr: Again, instead of commenting, it would be better to split this test in two tests

tick/optim/model/tests/linreg_with_intercepts_test.py
```diff
@@ -13,27 +13,82 @@
...
+        # Add gross outliers to the labels
+        y += intercept0
+        model = ModelLinRegWithIntercepts(fit_intercept=False).fit(X, y)
+        coeffs = np.random.randn(model.n_coeffs)
+        grad1 = model.grad(coeffs)
+        X2 = np.hstack((X, np.identity(n_samples)))
+        grad2 = X2.T.dot(X2.dot(coeffs) - y) / n_samples
+        np.testing.assert_almost_equal(grad1, grad2, decimal=10)
+
+        # *** Tests with global intercept ***
```
Mbompr: Same here

tick/optim/model/src/model_generalized_linear_with_intercepts.cpp
```diff
@@ -16,37 +16,41 @@ const char *ModelGeneralizedLinearWithIntercepts::get_class_name() const {
...
+  if (fit_intercept) {
+    ArrayDouble out_intercepts = view(out, n_features + 1, n_samples + n_features + 1);
+    if (fill) {
+      out_weights.mult_fill(x_i, alpha_i);
+      out_intercepts.fill(0);
+      out_intercepts[i] = alpha_i;
+      out[n_features] = alpha_i;
+    } else {
+      out_weights.mult_incr(x_i, alpha_i);
+      out_intercepts[i] = alpha_i;
```
Mbompr: I would either put `+=` or comment that this feature is the only one modifying this index.

PR #70, opened 2017-09-15
TICK-363 Improvement of the solvers docstrings

stephanegaiffas:


stephanegaiffas:
Ready for review @Mbompr : I let you update the doc for SCPG okay ? (using what I did for other batch solvers)

Issue #74, opened 2017-09-26
Set default values for base array 2d n_cols and n_rows


Mbompr:
At the moment the size of a BaseArray2d is correctly set to 0 in its default constructor but n_cols and n_rows remain unset which can lead to unexpected states, especially while trying to serialize an empty 2d array.

PR #75, opened 2017-09-26
#74 Set default values for base array 2d n_cols and n_rows

Mbompr:
Solves issue #74 to allow serialization of empty arrays


tick/base/array/tests/src/array_gtest.cpp
```diff
@@ -520,6 +520,28 @@ TYPED_TEST(ArrayTest, SortAbsTrackDecreasing) {
 namespace {
 
 template <typename ArchiveIn, typename ArchiveOut, typename TypeParam>
+void TestEmptySerialization() {
```
Mbompr: This test is nothing but `TestSerialization` for an empty array with tests on expected size.

PR #76, opened 2017-09-26
TICK-416 - Add support for mkn 

Dekken:
general compilation and C++ google-tests


README.md
```diff
@@ -107,6 +107,20 @@ This will build all extensions, and install them directly in the source director
...
+To use maiken, either download a provided [binary](https://github.com/dekken/maiken/tree/binaries) or compile the source following the instructions in the maiken README @ [Part 3: How to build Maiken]( )
+
+Once "mkn" is installed on your system, you should execture the shell script "./mkn.sh" to compile tick.
+
+As with tick in general, this requires python3 and all python pre-requisites. numpy/scipy/etc
+
+This method is provided mainly for developers to decrease compilation times. On windows this requires either [CYGWIN](https://www.cygwin.com/)/[MSYS](http://www.msys2.org/)
+
+To use [ccache](https://ccache.samba.org/) (or similar) with mkn, edit your ~/.maiken/settings.yaml file so it looks like one [here](https://github.com/Dekken/maiken/wiki/Alternative-configs) - this can greatly decrease repeated compilation times.
+
```
Mbompr: Don't you think we should move all of these in a wiki? 
Though, I don't know if versioning works with wiki.
(we can do this in another task).

mkn.yaml
```diff
@@ -0,0 +1,167 @@
...
+    src:
+      ./tick/random/src
+      ./tick/random/swig
+    install: ${lib_path}/random/build
+    out: crandom${lib_name}
+
+### See file ./sh/gtest/sh 
+  - name: gtest
+    parent: arg
+    main: tick/optim/model/tests/src/hawkes_models_gtest.cpp
```
Mbompr: Wouldn't this only launch one test file?

Mbompr: Should it be kept since `sh/gtest.sh` exists?

README.md
```diff
@@ -107,6 +107,20 @@ This will build all extensions, and install them directly in the source director
...
+To use maiken, either download a provided [binary](https://github.com/dekken/maiken/tree/binaries) or compile the source following the instructions in the maiken README @ [Part 3: How to build Maiken]( )
+
+Once "mkn" is installed on your system, you should execture the shell script "./mkn.sh" to compile tick.
+
+As with tick in general, this requires python3 and all python pre-requisites. numpy/scipy/etc
+
+This method is provided mainly for developers to decrease compilation times. On windows this requires either [CYGWIN](https://www.cygwin.com/)/[MSYS](http://www.msys2.org/)
+
+To use [ccache](https://ccache.samba.org/) (or similar) with mkn, edit your ~/.maiken/settings.yaml file so it looks like one [here](https://github.com/Dekken/maiken/wiki/Alternative-configs) - this can greatly decrease repeated compilation times.
```
Mbompr: We should mention that ccache is an external dependency and should be installed manually

sh/configure_env.sh
```diff
@@ -0,0 +1,106 @@
...
+#  $PCONF = "python-config" binary
+#  $LDARGS=Linking arguments
+#  $LIB_POSTFIX= python library format 
+#  $LIB_POSTEXT= system library file extention
+#  $PROFILES= array of mkn profiles to build
+# 
+######################################################################
+
+set -e
+
```
Mbompr: Maybe could you add comments such as
looking you python executable
looking for python-config, it will be needed to get Python link arguments

sh/configure_env.sh
```diff
@@ -0,0 +1,106 @@
...
+[ -z "$PY" ] && echo "python or python3 not on path, exiting with error" && exit 1
+
+[ -z "$PCONF"  ] && which python3-config &> /dev/null && PCONF="python3-config"
+[ -z "$PCONF"  ] && which python-config &> /dev/null && PCONF="python-config"
+[ -z "$PCONF"  ] && echo "python-config or python3-config not found on PATH, exiting with error" && exit 1;
+
+[ -z "$PINC" ] && for i in $($PCONF --includes); do PINC="${PINC}${i:2}:"; done
+[ -z "$PNIC" ] && PNIC=$($PY -c "import numpy as np; print(np.get_include())");
+
+LDARGS="$($PCONF --ldflags)"
```
Mbompr: Please explain why we need this sed :)

sh/configure_env.sh
```diff
@@ -0,0 +1,106 @@
...
+    Linux*)     LDARGS="${LDARGS} $(mkn -G nix_largs)";;
+    Darwin*)    LDARGS="${LDARGS} $(mkn -G bsd_largs)";;
+    CYGWIN*)    LDARGS="${LDARGS}";;
+    MINGW*)     LDARGS="${LDARGS}";;
+    *)          LDARGS="${LDARGS}";;
+esac
+[ ! -z "$LDFLAGS" ] && LDARGS="${LDARGS} ${LDFLAGS}"
+
+# IFS = Internal Field Separator - allows looping over lines with spaces
+IFS=$'\n'
```
Mbompr: Would this even work on Windows? I thought window was using \r\n but it's something I don't know much about.

sh/configure_env.sh
```diff
@@ -0,0 +1,106 @@
...
+
+LIB_POSTFIX=$($PY -c "import distutils; from distutils import sysconfig; print(sysconfig.get_config_var('SO'))")
+LIB_POSTEXT="${LIB_POSTFIX##*.}"
+LIB_POSTFIX="${LIB_POSTFIX%.*}"
```
Mbompr: What happens here?

mkn.yaml
Dekken: It's required by "sh/gtest.sh".

sh/configure_env.sh
Dekken: If it fails there is an error - which is the important part

sh/configure_env.sh
Dekken: Windows still requires MSYS or CYGWIN to use this, so it should be ok

sh/configure_env.sh
Dekken: In the file or here? 

sed is used there to remove parts of the string, if for instance we wish to override them with other values.

sh/configure_env.sh
Dekken: We extract the left of the file type, and the file extension type

```
LIB_POSTFIX=".cpython-35m-x86_64-linux-gnu.so"
LIB_POSTFIX="..cpython-35m-x86_64-linux-gnu"
LIB_POSTEXT="so"
```

mkn.yaml
Mbompr: Oh ok, so this is a default parameter. Maybe should we give example on how to use each .sh file?

sh/configure_env.sh
Mbompr: Yes I know since you were nect to me when you added it but I think it deserves a comment in the code, just like the above lines

sh/configure_env.sh
Mbompr: Oh yes right!

mkn.yaml
Dekken: It's explained at the top of the gtest.sh file - how it works, and how to use it - let me know if it's ok

sh/gtest.sh
```diff
@@ -0,0 +1,68 @@
...
+# Email : philip.deegan@polytechnique.edu 
+# Date  : 26 - September - 2017
+#
+# This script builds and launches google test files for tick
+#
+# Requires the mkn build tool
+#
+# Input arguments are optional but if used are to be the 
+#  test files to be compiled and run, otherwise all files
+#  with the sytanx "*gtest.cpp" are used
```
Mbompr: typo `sytanx `

mkn.yaml
Mbompr: Ok sorry, I'll give a more careful look when I will install it on my other machine

README.md
Dekken: I've no opinion on that, except if you put it on another source site (gitlab) the wiki won't go with the repo

sh/gtest.sh
Dekken: fixed

sh/configure_env.sh
Dekken: Comments added

sh/configure_env.sh
Dekken: Comments added

README.md
Dekken: Done

README.md
Dekken: Done

sh/configure_env.sh
```diff
@@ -0,0 +1,112 @@
...
+#  $PROFILES= array of mkn profiles to build
+# 
+######################################################################
+
+set -e
+
+# Finding the system python binary
+[ -z "$PY" ] && which python3 &> /dev/null && PY="python3"
+[ -z "$PY" ] && which python &> /dev/null && PY="python"
+[ -z "$PY" ] && echo "python or python3 not on path, exiting with error" && exit 1
```
Mbompr: We should check that python 3 is used. 
With python 2 I get the following, not so informative error
```
tick/base/array/src/alloc.h:20:3: error: use of undeclared identifier 'PyMem_RawFree'; did you mean 'PyMem_Free'?
  PyMem_RawFree(reinterpret_cast<void*>(ptr));
  ^~~~~~~~~~~~~
  PyMem_Free
```
Do we check anywhere that user has SWIG installed ?

Dekken: python3 is attempted to be used by default - it can be overridden by executing

```
export PY=/path/to/python3
```

Similarly for PCONF

```
export PCONF=/path/to/python-config
```

in "mkn.sh" we attempt to use swig. so it will fail if it not found - swig is mentioned in the setup requirements in the README

Dekken: I've added checks for

- to see if swig is on path (overridable variable $SWIG)
- that python found is version 3

Issue #82, opened 2017-09-27
Master FAIL when using Xcode 9 (updated yesterday) on MacOS -Wunused-private-field


stephanegaiffas:
Since Xcode 9 (updated yesterday) -Wunused-private-field leads to compilation fail on MacOS (on my laptop), see below

There's a problem with CEREAL :

compile options: ‘-DSCIPY_MKL_H -DHAVE_CBLAS -DTICK_CBLAS_AVAILABLE -Itick/third_party/cereal/include -I/Users/stephane.gaiffas/anaconda/lib/python3.5/site-packages/numpy/core/include -I/Users/stephane.gaiffas/anaconda/include/python3.5m -c’
extra options: ‘-D_FILE_OFFSET_BITS=64 -DPYTHON_LINK -DNPY_NO_DEPRECATED_API=NPY_1_7_API_VERSION -I./tick/base/src/ -DTICK_SPARSE_INDICES_INT32 -std=c++11 -O3 -ffast-math -Wall -Werror -Wno-uninitialized -Itick/base/src/ -Itick/base/swig/ -Itick/base/array/src/ -Itick/base/array/swig/ -DDEBUG_COSTLY_THROW’
gcc: tick/base/src/math/t2exp.cpp
gcc: tick/base/src/time_func.cpp
gcc: tick/base/swig/base_module_wrap.cpp
In file included from tick/base/swig/base_module_wrap.cpp:3141:
In file included from ./tick/base/src/serialization.h:7:
tick/third_party/cereal/include/cereal/archives/json.hpp:529:42: error: private field ‘itsValueItEnd’ is not used [-Werror,-Wunused-private-field]
          ValueIterator itsValueItBegin, itsValueItEnd;    //!< The value iterator (array)
                                         ^
1 error generated.
In file included from tick/base/swig/base_module_wrap.cpp:3141:
In file included from ./tick/base/src/serialization.h:7:
tick/third_party/cereal/include/cereal/archives/json.hpp:529:42: error: private field ‘itsValueItEnd’ is not used [-Werror,-Wunused-private-field]
          ValueIterator itsValueItBegin, itsValueItEnd;    //!< The value iterator (array)
                                         ^
1 error generated.
[...]

PR #86, opened 2017-09-28
#82 Temporary fix: Removing -Werror

Mbompr:


Mbompr:
This was already merged but with pull-merge commits so I've cleaned the whole thing a bit.

Issue #84, opened 2017-09-27
Fixing problems in documentation


stephanegaiffas:
Removed references to non-existing tick.plot.stem and tick.metrics.{fdp, recall} ane removed verbosity from code generating plots in doc and some errors

PR #87, opened 2017-09-28
Removed references to non-existing tick.plot.stem and tick.metrics.{f…

Mbompr:
…dp, recall} ane removed verbosity from code generating plots in doc and some errors

Solves #84


Mbompr:
This was already merged but with pull-merge commits so I've cleaned the whole thing a bit.

Issue #93, opened 2017-10-05
Windows build with mkn


Dekken:
Only distutils works properly on windows so far. 

Figure out how to replicate

PR #103, opened 2017-10-06
#93 - mkn working on windows with MSVC

Dekken:
For https://github.com/X-DataInitiative/tick/issues/93

### DIFF
--- a/mkn.yaml
+++ b/mkn.yaml
@@ -20,7 +20,7 @@ property:
            -D_FILE_OFFSET_BITS=64 -DPYTHON_LINK
            -DNDEBUG
     win_cargs:
-      -GL -MD -EHsc -Zc:inline -DBUILDING_DLL -O2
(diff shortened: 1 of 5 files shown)


mkn.yaml
```diff
@@ -20,7 +20,9 @@ property:
            -D_FILE_OFFSET_BITS=64 -DPYTHON_LINK
            -DNDEBUG
     win_cargs:
-      -GL -MD -EHsc -Zc:inline -DBUILDING_DLL -O2
+      -Ox -W3 -GL -DNDEBUG -MT -DDEBUG_COSTLY_THROW
```
Mbompr: `-DDEBUG_COSTLY_THROW` is only on windows but not on other platforms

Dekken:
Yeah I removed it - if you set "DEBUG=1" on command line it will add it anyhow

Issue #73, opened 2017-09-26
Add Kromagnon algorithm aka Asynchronous SVRG


Mbompr:
This can start once #61 is merged

The idea is to run in parallel computations in method `solve_sparse_proba_updates `

Paper: https://arxiv.org/pdf/1507.06970.pdf

Dekken:
Re worked what I had into something usable - will PR once cleaned up

PR #107, opened 2017-10-10
#73 Add ASVRG - Asynchronous SVRG - gives SVRG a multithreaded option

Dekken:
https://github.com/X-DataInitiative/tick/issues/73

examples/plot_asynchronous_stochastic_solver.py - courtesy of @Mbompr

### DIFF
--- a/tick/optim/solver/src/svrg.cpp
+++ b/tick/optim/solver/src/svrg.cpp
@@ -8,9 +8,11 @@ SVRG::SVRG(ulong epoch_size,
            RandType rand_type,
            double step,
            int seed,
+           int n_threads,
            VarianceReductionMethod variance_reduction)
     : StoSolver(epoch_size, tol, rand_type, seed),
-      step(step), variance_reduction(variance_reduction),
+      n_threads(n_threads), step(step),
+      variance_reduction(variance_reduction),
       ready_step_corrections(false) {}
 
 void SVRG::set_model(ModelPtr model) {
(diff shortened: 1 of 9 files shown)


Dekken:
This is not atomic - it's possible we should consider implementing an atomic version - I imagine this would require a change to the array module

tick/optim/solver/svrg.py
```diff
@@ -77,7 +77,10 @@ def __init__(self, step: float = None, epoch_size: int = None,
                  rand_type: str = 'unif', tol: float = 0.,
                  max_iter: int = 100, verbose: bool = True,
                  print_every: int = 10, record_every: int = 1,
-                 seed: int = -1, variance_reduction: str = 'last'):
+                 seed: int = -1, variance_reduction: str = 'last',
+                 threads: int = 1):
```
Mbompr: To be consistent with the library this parameter should be called `n_threads`

tick/optim/solver/src/svrg.h
```diff
@@ -41,12 +42,25 @@ class SVRG : public StoSolver {
...
+      const bool use_intercept,
+      ProxSeparable*& casted_prox);
+
  public:
   SVRG(ulong epoch_size,
        double tol,
        RandType rand_type,
        double step,
        int seed = -1,
+       int threads = 1,
```
Mbompr: To be consistent with the library this parameter should be called `n_threads`

tick/optim/solver/src/svrg.h
```diff
@@ -17,6 +17,7 @@ class SVRG : public StoSolver {
   };
 
  private:
+  int threads = 1;
```
Mbompr: To be consistent with the library this parameter should be called `n_threads`

tick/optim/solver/src/svrg.cpp
```diff
@@ -68,19 +70,24 @@ void SVRG::compute_step_corrections() {
...
-    for (ulong j = 0; j < iterate.size(); ++j) {
-      iterate[j] = iterate[j] - step * (grad_i[j] - grad_i_fixed_w[j] + full_gradient[j]);
-    }
-    prox->call(iterate, step, iterate);
-    if (variance_reduction == VarianceReductionMethod::Random && t == rand_index) {
-      next_iterate = iterate;
-    }
-    if (variance_reduction == VarianceReductionMethod::Average) {
-      next_iterate.mult_incr(iterate, 1.0 / epoch_size);
+  if (threads > 1) {
```
Mbompr: Just wondering, don't you think we could isolate this code in a function and pass `dense_single_thread_solver` / `sparse_single_thread_solver` as argument? It would avoid code duplication and could afterwards probably be shared with all sto solvers

tick/base/__init__.py
```diff
@@ -10,10 +10,14 @@ def _set_mpl_backend():
...
-                 and matplotlib.get_backend() == 'MacOSX'
-                 )):
+    if (( os.name == 'posix' and 'DISPLAY' not in os.environ
+          and not
+            (sys.platform == 'darwin'
+             and matplotlib.get_backend() == 'MacOSX')
+        )
+          or ('DISPLAY' in os.environ
+             and os.environ['DISPLAY'] == '-1')
+        ):
```
Mbompr: Maybe this addition should be in another PR describing clearly why it is needed?

tick/optim/solver/src/svrg.h
```diff
@@ -41,12 +42,25 @@ class SVRG : public StoSolver {
 
   void compute_step_corrections();
 
+  void dense_single_thread_solver(const ulong& next_i);
+
+  // ProxSeparable* is a raw pointer here as the
+  //  ownership of the pointer is handled by
+  //  a shared_ptr which is above it in the same
+  //  scope so a shared_ptr is not needed
```
Mbompr: Thanks for this comment!

tick/base/__init__.py
Dekken: Oh that's what you meant by "isolate" sure that's cool

tick/optim/solver/src/svrg.cpp
Dekken: can we put that in a different ticket - I prefer simplicity but that the method parameters are not the same it introduces difficulty

Mbompr: Makes sense, it will mostly be useful once we generalize asynchronous solving to all stochastic solvers.

Issue #104, opened 2017-10-09
Experience with Anaconda - Python 3.6.2 on windows


Courvoisier13:
Hi guys, here is my experience with Anaconda - Python 3.6.2 on Windws 10 / Windows 7 64 bits .
I hope it will help improve the clarity for other people.

1. Add `blas_info = {}` to `setup.py ` line 107
2. The wheels and the and dependencies are not needed (included in anaconda) but it doesn't hurt to check. 
3. SWIG: uses version > 3.07 as indicated (version 2 would not work). I used SWIG 3.0.12
    3.a. Windows 10: ADD the SWIG.exe version 3.0.7 file to the "Path" environment variable.
    3.b. Windows 7: ADD the SWIG **folder** version 3.0.7 file to the "Path" environment variable. swig.exe would not work in Windows 7
4. the command `/path/to/VS/VC/Auxiliary/Build/vcvarsall.bat amd64` is not clear. This is actually the `vcvarsall.bat` in the VS BUILD TOOLS requested above. It is located in `C:\Program Files (x86)\Microsoft Visual Studio\2017\BuildTools\VC\Auxiliary\Build` by default.
5. There is also the two SWIG environment variables ([see](http://www.swig.org/Doc3.0/Windows.html#Windows_examples)). I didn't try without those, but it's possible it would work without.
PYTHON_INCLUDE: E:\anaconda3\include
PYTHON_LIB: E:\anaconda3\libs\python36.lib

Then the compilation should work.

Dekken:
Hi Courvoisier13,

Thanks for the information.

1. This is already [there](https://github.com/X-DataInitiative/tick/blob/master/setup.py#L85) - what you probably experiencing is a failure to find an appropriate blas with some warnings - yes?

2. That makes sense.

3. The version of swig is a copy paste error - and yes there is no bin directory for these, rather the root install can be considered the bin directory.

4. I will add a reference to the default, however it is generally wise to avoid the default install location (in my mind) as it contains spaces which are anathema when dealing with command line tooling.

I will handle these in conjunction with https://github.com/X-DataInitiative/tick/issues/105

Dekken:
@Mbompr  has reminded me of a problem he was having regarding blas_info, and this is possibly what is occuring for you - this doesn't happen for my windows environment so I find it a bit peculiar.

What seems to be happening is that the -D define arguments for BLAS are added to the compile commands, but no blas exists and thus the head files cannot be found.

To be investigated further.

[...]

PR #113, opened 2017-10-12
#104 Anaconda on windows 

Dekken:
resolve https://github.com/X-DataInitiative/tick/issues/104 part 1

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -93,7 +93,17 @@ try:
         numpy_include = np.get_numpy_include()
 
     # Determine if we have an available BLAS implementation
-    if 'bdist_wheel' not in sys.argv:
+    if platform.system() == 'Windows':
+        try:
+            with open(os.devnull, 'w') as devnull:
+                exitCode = subprocess.check_output(
+                    "python tools/python/blas/check_cblas.py build_ext",
+                    stderr=devnull,
+                    shell=True)
+                blas_info = get_info("blas_opt", 0)
+        except subprocess.CalledProcessError as subError:
+            print("Error executing check_cblas.py - cblas not found")
+    elif 'bdist_wheel' not in sys.argv:
         blas_info = get_info("blas_opt", 0)
 
     numpy_available = True
--- a/sh/configure_env.sh
+++ b/sh/configure_env.sh
@@ -161,15 +161,48 @@ LDARGS=$(echo "$LDARGS" | sed -e "s/ CoreFoundation//g")
 # IFS = Internal Field Separator - allows looping over lines with spaces
 IFS=$'\n'
 
-
+ANACONDA=0
+LIBLDARGS=""
 # Add library path if using anaconda python
 PYVER=$($PY --version 2>&1 )
 if [[ "$PYVER" == *"Anaconda"* ]]; then
+  ANACONDA=1
   ANACONDA_LIB_PATH=$(linkread $(dirname $(which $PY))/../lib)
-  LDARGS="-L${ANACONDA_LIB_PATH} -Wl,-rpath,${ANACONDA_LIB_PATH} $LDARGS -lmkl_rt -lpthread"
-  LDARGS="-dynamiclib -undefined dynamic_lookup -Wl,-headerpad_max_install_names"
-  CXXFLAGS="-DSCIPY_MKL_H -DHAVE_CBLAS -DTICK_CBLAS_AVAILABLE $CXXFLAGS"
+  if [[ "$unameOut" == "Darwin"* ]]; then
+    # Unknown reason why these are not required for Anaconda -
+    #  Not just that but neither is the -lpython
+    #  The line below is left intentionally commented as a reference of what should be linked
+    LDARGS="" #-L${ANACONDA_LIB_PATH} -Wl,-rpath,${ANACONDA_LIB_PATH} $LDARGS -lmkl_rt -lpthread"
+    LIBLDARGS="-dynamiclib -undefined dynamic_lookup -Wl,-headerpad_max_install_names"
+    CXXFLAGS="-DSCIPY_MKL_H -DHAVE_CBLAS -DTICK_CBLAS_AVAILABLE $CXXFLAGS"
+  fi
 fi
+
+
+##
+# MKN_P_ARRAY exists to allow platfrom 
+#  specific properties to be passed to mkn if required
+MKN_P_ARRAY=(lib_name=$LIB_POSTFIX)
+case "${unameOut}" in
+    Linux*)     LDARGS="${LDARGS}";;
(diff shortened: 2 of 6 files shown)


setup.py
```diff
@@ -93,7 +93,17 @@
...
 
     # Determine if we have an available BLAS implementation
-    if 'bdist_wheel' not in sys.argv:
+    if platform.system() == 'Windows':
+        try:
+            with open(os.devnull, 'w') as devnull:
+                exitCode = subprocess.check_output(
+                    "python tools/python/blas/check_cblas.py build_ext",
+                    stderr=devnull,
+                    shell=True)
```
Mbompr: Add a comment: 
If blas is found we will be able to link tick against it

sh/configure_env.sh
```diff
@@ -161,16 +161,38 @@ LDARGS=$(echo "$LDARGS" | sed -e "s/ CoreFoundation//g")
...
 if [[ "$PYVER" == *"Anaconda"* ]]; then
+  ANACONDA=1
   ANACONDA_LIB_PATH=$(linkread $(dirname $(which $PY))/../lib)
-  LDARGS="-L${ANACONDA_LIB_PATH} -Wl,-rpath,${ANACONDA_LIB_PATH} $LDARGS -lmkl_rt -lpthread"
-  LDARGS="-dynamiclib -undefined dynamic_lookup -Wl,-headerpad_max_install_names"
-  CXXFLAGS="-DSCIPY_MKL_H -DHAVE_CBLAS -DTICK_CBLAS_AVAILABLE $CXXFLAGS"
+  if [[ "$unameOut" == "Darwin"* ]]; then
+    # Unknown reason why these are not required for Anaconda -
+    #  Not just that but neither is the -lpython
+    LDARGS="" #-L${ANACONDA_LIB_PATH} -Wl,-rpath,${ANACONDA_LIB_PATH} $LDARGS -lmkl_rt -lpthread"
```
Mbompr: If this a left comment?

sh/configure_env.sh
```diff
@@ -161,16 +161,38 @@ LDARGS=$(echo "$LDARGS" | sed -e "s/ CoreFoundation//g")
...
 fi
 
+MKN_P_ARRAY=(lib_name=$LIB_POSTFIX)
+case "${unameOut}" in
+    Linux*)     LDARGS="${LDARGS}";;
+    Darwin*)    LDARGS="${LDARGS}";;
+    CYGWIN*)    LDARGS="${LDARGS}";;
+    MINGW*)     LDARGS="${LDARGS}";;
+    *)          LDARGS="${LDARGS}";;
+esac
```
Mbompr: does this case do anything more than `LDARGS="${LDARGS}"`?

tools/python/blas/check_cblas.py
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env python
+# -*- coding: utf8 -*-
+
+# python setup.py build_ext --inplace
+
+"""
+setup.py file
```
Mbompr: Maybe can we explain here that this file is here to see if we can link against blas

setup.py
```diff
@@ -93,7 +93,17 @@
         numpy_include = np.get_numpy_include()
 
     # Determine if we have an available BLAS implementation
-    if 'bdist_wheel' not in sys.argv:
+    if platform.system() == 'Windows':
```
Mbompr: Don't you think we should do this on all platforms even if the bug currently only occurs on Windows?

sh/configure_env.sh
```diff
@@ -161,16 +161,38 @@ LDARGS=$(echo "$LDARGS" | sed -e "s/ CoreFoundation//g")
...
+MKN_P=""
```
Mbompr: Why don't we call it `MKN_PROPERTIES` ? It looks easier to understand this way

setup.py
Dekken: finding BLAS seems to work ok on Linux - i.e. it only adds the defines when there's one installed, at least in my experience

setup.py
Dekken: good question - I've never installed BLAS on windows so I have no idea.

If they threw the blas.lib file into the python lib directory then sure that could work

sh/configure_env.sh
Dekken: yes - LDARGS should include the ANACONDA_LIB_PATH and all the rest but it doesn't work when it's done so TBC

setup.py
Mbompr: I agree but if for any reason `check_cblas.py` failed, wouldn't this mean that we should set `blas_info` to `{}`, no matter on which platform this happens?

sh/configure_env.sh
Dekken: This is a place holder for the "MKN_P_ARRAY" in case we wish to specificy platform specific properties

sh/configure_env.sh
Dekken: the "-P" argument for mkn refers to "properties

tools/python/blas/check_cblas.py
Dekken: sure

sh/configure_env.sh
Mbompr: Oh ok, this is what we would have expected to work but chich doesn't?

sh/configure_env.sh
Mbompr: Oh, ok maybe it's worth leaving a tiny comment just before defining `MKN_P_ARRAY` :)

sh/configure_env.sh
Dekken: yes indeed

it's strange too as if you check with otool the "mkl_rt" library isn't referenced by any library when built with distutils - so it might be used but ignored

setup.py
Dekken: ```check_cblas.py``` is only a compilation check, and distutils might do some other magic to add blas arguments to the compile/link commands

https://github.com/X-DataInitiative/tick/blob/master/setup.py#L97 

if we see issues on other platforms it's not difficult to include the compilation there too

sh/configure_env.sh
Dekken: sure

setup.py
Mbompr: Alright, I was not thinking about distutils magic, I thought `check_cblas.py` would necessarily imply setup.py to fail

sh/configure_env.sh
Dekken: https://github.com/X-DataInitiative/tick/pull/113/files#diff-8c1ca1a035c8f4df5bca4bc9814477f9R200

PR #126, opened 2017-10-27
Release 0.3.0

Mbompr:


Mbompr:
Well, the 3rd digit is for a patch if necessary and the fourth is because if anything was wrong while uploading to pip we cannot submit the same version number again.

Issue #124, opened 2017-10-26
Grand flattening and reorganization of `tick`


stephanegaiffas:
# Grand reorganization of `tick`

A lot of features have been added and now `tick` is much more than a library that fits point processes.
The main issue is that the `tick.inference` module contains a lot of methods that have nothing to do together. 
I suggest that we reorganize the things in big **families** of methods, putting together inference, simulation and models together in such families.

This is a big change, that will pose problems with C++ includes. Several base class must be included in a `tick.base` module as soon as they are used in several modules. An example is `ModelFirstOrder` python class, or the `Model` C++ class, that are parent of almost all models in `tick`. Such classes will have to be in a `tick.base`.

	> TODO: list of all the things that go in `tick.base` and `tick.*.base` modules.

Here is a first attempt of grand reorganization of `tick`. I've also renamed a lot of classes, so that inference and simulation classes names match...

	> TODO: Discuss and give your opinion about this here

## 1. `tick.linear_model`

Inference, simulation and models all belong to a common namespace

- `tick.linear_model.LinearRegression`
- `tick.linear_model.LogisticRegression`
- `tick.linear_model.PoissonRegression`
- `tick.linear_model.SimuLinearRegression`
- `tick.linear_model.SimuLogisticRegression`
- `tick.linear_model.SimuPoissonRegression`
- `tick.linear_model.ModelLeastSquares`
- `tick.linear_model.ModelLogistic`
- `tick.linear_model.ModelHinge`
- `tick.linear_model.ModelSmoothedHinge`
- `tick.linear_model.ModelQuadraticHinge`
- `tick.linear_model.ModelPoisson`

## 2. `tick.survival`

- `tick.survival.nelson_aalen`
- `tick.survival.kaplan_meier`
- `tick.survival.CoxRegression`
- `tick.survival.SimuCoxRegression`
- `tick.survival.ModelCoxRegPartialLik`
- `tick.survival.ModelSCCS`

We will put extra things here : Cox regression with the full likelihood, the SCCS model from Maryan, etc.

## 3. `tick.point_process`

- `tick.point_process.HawkesExpKern`
- `tick.point_process.HawkesSumExpKern`
- `tick.point_process.HawkesEM`
- `tick.point_process.HawkesADM4`
- `tick.point_process.HawkesBasisKernels`
- `tick.point_process.HawkesSumGaussians`
- `tick.point_process.HawkesConditionalLaw`
- `tick.point_process.SimuPoissonProcess`
[...]

PR #127, opened 2017-11-02
restructure

Dekken:
For https://github.com/X-DataInitiative/tick/issues/124

- Separate sources and includes
- Single include directory
- All include paths complete and relative from ./include
- Minimise setup.py - only use search directories
- Move array and array_test up one level from under "base"
- CMake files updated to pass travis build
- all sh files in root moved to ./sh - updated accordingly


.gitmodules
```diff
@@ -1,3 +1,3 @@
 [submodule "tick/third_party/cereal"]
```
Mbompr: not sure but shouldn't this be lib/third_party/cereal ?

lib/include/tick/array/carray_python.h
```diff
@@ -18,7 +18,7 @@
 #include "ssparsearray2d.h"
 #include "sbasearray.h"
 #include "sbasearray2d.h"
-#include "varray.h"
+#include "tick/array/varray.h"
```
Mbompr: When do we need full path and when can we use file name directly?

clean_build_test.sh
```diff
@@ -1,6 +0,0 @@
-#!/usr/bin/env bash
-
-echo "CLEAN"
-./full_clean.sh
-
```
Mbompr: I think it was cool to have this file even if its content only executes python commands. It is useful when people have problem with their local installation.

lib/src/cpp/array/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
+add_library(tick_array EXCLUDE_FROM_ALL
```
Mbompr: Can we define at the top of this file:
set(ARRAY_INCLUDE_DIR, "${PROJECT_SOURCE_DIR}/lib/include/tick/array/")
?

lib/src/swig/CMakeLists.txt
```diff
@@ -21,8 +21,8 @@ endif()
```
Mbompr: Why is the main CMakeLists in the swig folder?

sh/swig.sh
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/usr/bin/env bash-Iinclude
```
Mbompr: typo?

tick/base/__init__.py
```diff
@@ -18,7 +18,7 @@ def _set_mpl_backend():
 
 _set_mpl_backend()
 
-from .array import *
+from tick.array import *
```
Mbompr: Why not but was this necessary?

tick/plot/tests/plot_hawkes_test.py
```diff
@@ -1,67 +0,0 @@
-# License: BSD 3 clause
```
Mbompr: You have removed test folders for plot

.gitmodules
Dekken: probably, not sure it matters but I'll change it for consistency

clean_build_test.sh
Dekken: so what, remove it?

lib/src/cpp/array/CMakeLists.txt
Dekken: we can - you're the CMake guy! ^_^

lib/src/swig/CMakeLists.txt
Dekken: Copy paste - will remove

sh/swig.sh
Dekken: :|


yes

tick/base/__init__.py
Dekken: array is no longer a subdirectory of base so I believe so

tick/plot/tests/plot_hawkes_test.py
Dekken: whoops

Mbompr:
Warning each time we run setup.py
UserWarning: Included file t2exp.inl in folder lib/src/cpp/base/math has an unknown extension ".inl"
Looks like this file extension is not handled

Dekken:
https://github.com/X-DataInitiative/tick/blob/master/setup.py#L256

```
                if ext == '.inl':
                    pass
```

lib/src/cpp/array/CMakeLists.txt
Dekken: Done
https://github.com/Dekken/tick/blob/re-org/lib/src/cpp/array/CMakeLists.txt

lib/include/tick/array/alloc.h
```diff
@@ -5,7 +5,7 @@
 
 /** @file */
 
-#include "debug.h"
+#include "tick/base/debug.h"
```
Mbompr: is full path necessary or a good practice or a find/replace mistake?

lib/src/cpp/array/CMakeLists.txt
Mbompr: I've done it for other CMakeLists files as well

lib/src/cpp/base/math/t2exp.inl
```diff
@@ -0,0 +1,2078 @@
+/*
```
Mbompr: This file exists at two locations:
* lib/src/cpp/base/math/t2exp.inl
* lib/include/tick/base/math/t2exp.inl
One should probably be removed

tick/inference/inference.py
```diff
@@ -0,0 +1,23 @@
+# License: BSD 3 clause
```
Mbompr: This file is a duplication of tick/inference/tests/inference.py and should be deleted

tick/optim/model/generalized_linear_model.py
```diff
@@ -0,0 +1,70 @@
+# License: BSD 3 clause
```
Mbompr: This file is a duplication of  tick/optim/model/tests/generalized_linear_model.py and should be deleted

tick/optim/model/hawkes_utils.py
```diff
@@ -0,0 +1,111 @@
+# License: BSD 3 clause
```
Mbompr: This file is a duplication of tick/optim/model/hawkes_utils.py and should be deleted

lib/include/tick/array/alloc.h
Mbompr: Ok now full path must be used everywhere, the only exception is when two .h files are in the same folder.

Dekken:
Auto closed on commit?

Issue #133, opened 2017-11-21
Methods needed to assess fit of Hawkes Process


sharonxu:
I think it would be extremely helpful to have methods to assess the fit of a Hawkes Process, namely:

1. Residuals. To assess fit, we would like to check that the residual process is homogenous and has exponentially distributed inter-event times (F. Lorenzen: Analysis of Order Clustering Using High Frequency Data: A Point Process Approach [pdf](http://arno.uvt.nl/show.cgi?fid=126950).

2. Conditional intensity function given the history/timestamped data (+1 for plotting functionality with respect to the corresponding timestamped data).

Mbompr:
Hi, indeed it looks pretty interesting, we simply did not have enough time for that yet. I can help you if you want to add one of these features, otherwise, I might implement them but maybe not in a very near future.

stephanegaiffas:
This is certainly interesting, let's schedule this for a milestone Version 0.5

majkee15:
Hi, I have implemented a workaround to asses the quality of the fits. Not perfect, definitely slow but should work. See: https://github.com/majkee15/TICK-goodness-of-fit

mvicrob:
Thanks @majkee15, I just used it on my 1d data. It seems to work fine.

Dekken:
@majkee15 / @mvicrob 

Would either of you be able to provide me with a small script which tests this so I can see about adding it

mvicrob:
In fact, the notebook provided by @majkee15 is probably what you are looking for https://github.com/majkee15/TICK-goodness-of-fit/blob/master/time_transform_with_tick.ipynb.

I slightly modified this notebook for my need:

- For a 1d Hawkes process (I'm interested in an intertrade time series)
- With Statsmodel QQ plot instead of Scipy Probplot (because of https://stackoverflow.com/questions/48108582/how-to-interpret-scipy-stats-probplot-results)
- With a loop on potential decays, as in the example of https://stats.stackexchange.com/questions/366381/fit-hawkes-process-to-1d-data-using-python-package-tick

Below if a very draft idea of the result (tested with Python 3.6):

```
import matplotlib.pyplot as plt
import numpy as np
from tick.hawkes import SimuHawkesExpKernels, HawkesExpKern
from tick.plot import plot_point_process
from scipy import integrate
from scipy import stats
from tick.plot import plot_basis_kernels, plot_hawkes_kernels
import statsmodels.api as sm

plt.style.use('ggplot')


[...]

PR #136, opened 2017-11-27
Plot Hawkes process estimated intensity

Mbompr:
Do half of #133 

Here is the plot generated by the example.
@sharonxu  Do you have any suggestion?

estimated_intensity


sharonxu:
Hey @Mbompr thanks! This is exactly what I had in mind, it looks great.

Issue #137, opened 2017-11-30
Allow for serializing/deserialzing SparseArray/2d


Dekken:
This is to help with C++ tests/benchmarks

PR #138, opened 2017-11-30
Allow for serializing/deserialzing SparseArray2d

Dekken:
Referencing #137

### DIFF
--- a/lib/include/tick/array/serializer.h
+++ b/lib/include/tick/array/serializer.h
@@ -1,0 +1,59 @@
+#ifndef TICK_BASE_ARRAY_SRC_SERIALIZER_H_
+#define TICK_BASE_ARRAY_SRC_SERIALIZER_H_
+
+#include "array.h"
+#include "sarray.h"
+#include "array2d.h"
+#include "sarray2d.h"
+
+#include "sparsearray2d.h"
+#include "ssparsearray2d.h"
+
+#include <fstream>
+#include <cereal/archives/portable_binary.hpp>
+
+template<typename A>
+void array_to_file(std::string _file, const A &array) {
+  std::ofstream ss(_file, std::ios::out | std::ios::binary);
+  {
+    cereal::PortableBinaryOutputArchive oarchive(ss);
+    oarchive(array);
+  }
+}
+
+template<typename A>
+std::shared_ptr<A> array_from_file(const std::string &_file) {
+  auto arr = std::make_shared<A>();
+  {
+    std::ifstream data(_file, std::ios::in | std::ios::binary);
+    cereal::PortableBinaryInputArchive iarchive(data);
+    iarchive(*arr.get());
+  }
+  return arr;
+}
+
+void tick_double_array_to_file(std::string _file, const ArrayDouble &array) {
+  array_to_file<ArrayDouble>(_file, array);
+}
+
+SArrayDoublePtr tick_double_array_from_file(std::string _file) {
+  return array_from_file<SArrayDouble>(_file);
+}
+
+void tick_double_array2d_to_file(std::string _file, const ArrayDouble2d &array) {
+  array_to_file<ArrayDouble2d>(_file, array);
+}
+
+SArrayDouble2dPtr tick_double_array2d_from_file(std::string _file) {
+  return array_from_file<SArrayDouble2d>(_file);
+}
+
+void tick_double_sparse2d_to_file(std::string _file, const SparseArrayDouble2d &array) {
+  array_to_file<SparseArrayDouble2d>(_file, array);;
+}
+
+SSparseArrayDouble2dPtr tick_double_sparse2d_from_file(std::string _file) {
+  return array_from_file<SSparseArrayDouble2d>(_file);
+}
+
+#endif  // TICK_BASE_ARRAY_SRC_SERIALIZER_H_
--- a/lib/include/tick/array/sparsearray2d.h
+++ b/lib/include/tick/array/sparsearray2d.h
@@ -12,6 +12,13 @@
 template<typename T>
 class SSparseArray2d;
 
+// Class forwarding to allow "friend class" declarations
+namespace cereal
+{
+  template <typename T, class Archive> void save(Archive& ar, const SparseArray2d<T>& s);
+  template <typename T, class Archive> void load(Archive& ar, SparseArray2d<T>& s);
+}
+
 /*! \class SparseArray2d
  * \brief Template class for basic sparse 2d-arrays of type `T`.
  * It is stored in Compressed Sparse Row matrix form :
@@ -34,6 +41,9 @@ class SSparseArray2d;
  */
 template<typename T>
 class SparseArray2d : public BaseArray2d<T> {
+ // "friend class" declarations to allow private member access when de/serializing
+ template <typename T1, class Archive> friend void cereal::save(Archive& ar, const SparseArray2d<T1>& s);
+ template <typename T1, class Archive> friend void cereal::load(Archive& ar, SparseArray2d<T1>& s);
  protected:
     using BaseArray2d<T>::_size;
     using BaseArray2d<T>::_size_sparse;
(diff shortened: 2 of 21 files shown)


lib/include/tick/array/sparsearray2d.h
```diff
@@ -89,6 +102,43 @@ class SparseArray2d : public BaseArray2d<T> {
...
+      return spd2d;
+    }
+
+    static void TO_FILE(std::string _file, const SparseArray2d<T>& array){
+      std::ofstream ss(_file, std::ios::out | std::ios::binary);
+      {
+        cereal::PortableBinaryOutputArchive oarchive(ss);
+        oarchive(array);
+      }
+      /** Code to compare serialized versions to original
```
Mbompr: I Think this should go in a unit test rather left commented here.

lib/include/tick/array/sparsearray2d.h
```diff
@@ -135,6 +185,62 @@ Array2d<T> BaseArray2d<T>::as_array2d() {
     return c;
 }
 
+// Using save/load methods directory on SparseArray2d<T> does not compile simply
```
Mbompr: What does this mean?

lib/include/tick/array/sparsearray2d.h
```diff
@@ -135,6 +185,62 @@ Array2d<T> BaseArray2d<T>::as_array2d() {
...
+      s._data        = s_data;
+      s._indices     = s_indices;
+      s._row_indices = s_row_indices;
+
+      s.is_data_allocation_owned = 1;
+      s.is_indices_allocation_owned = 1;
+      s.is_row_indices_allocation_owned = 1;
+
+    }catch(const std::exception& e){
+      std::cerr << e.what() << std::endl;
```
Mbompr: maybe should we throw an exception? For example with TICK_ERROR macro

lib/include/tick/array/ssparsearray2d.h
```diff
@@ -277,7 +277,8 @@ std::shared_ptr<SSparseArray2d<T>> SSparseArray2d<T>::new_ptr(ulong n_rows, ulon
 // Constructor from a SparseArray2d (copy is made)
 template<typename T>
 std::shared_ptr<SSparseArray2d<T>> SSparseArray2d<T>::new_ptr(SparseArray2d<T> &a) {
-    std::shared_ptr<SSparseArray2d<T>> aptr = SSparseArray2d<T>::new_ptr(a.n_rows(), a.n_rows(), a.size_sparse());
+    std::shared_ptr<SSparseArray2d<T>> aptr = SSparseArray2d<T>::new_ptr(a.n_rows(), a.n_cols(), a.size_sparse());
+    aptr->_size = a.size();
```
Mbompr: Nice catch!

tick/array/tests/serializer.py
```diff
@@ -0,0 +1,43 @@
...
+
+from tick.array.build.array import Serializer
+from tick.optim.solver.tests.solver import TestSolver
+
+from tick.optim.model import ModelLinReg, ModelLogReg, ModelPoisReg, ModelCoxRegPartialLik
+from tick.simulation import SimuLinReg, SimuLogReg, SimuPoisReg, SimuCoxReg, weights_sparse_gauss
+
+from scipy.sparse import csr_matrix
+
+model_simu_map = {
```
Mbompr: it would be cleaner if these instantiations were in the setup method of test?

tools/python/serializer.py
```diff
@@ -0,0 +1,28 @@
+# License: BSD 3 clause
```
Mbompr: I would rather put this kind of utility closer to the fetch_tick_dataset method. I also think that it should be a method rather than a script.

tools/travis/docker_run.sh
```diff
@@ -11,6 +11,12 @@ eval "$(pyenv init -)"
 pyenv global ${PYVER}
 pyenv local ${PYVER}
 
+gcc -v
+
+git log -10
+
+swig -version
```
Mbompr: To be removed ?

tools/travis/osx_run.sh
```diff
@@ -2,12 +2,22 @@
...
 set -e -x
 
-shell_session_update() { :; }
+shell_session_update() { 
+  set +e
+  echo "This is so travis doesn't break"
+  return 0;
+}
+
+rvm get head
```
Mbompr: I think that everything that is related to Travis compilation should be done in another PR.

lib/include/tick/array/serializer.h
```diff
@@ -0,0 +1,20 @@
+
+
+#include "sparsearray.h"
+#include "sparsearray2d.h"
+
+class Serializer{
```
Mbompr: Do we really still need this class? As it has no state we should be able to expose simply array_to_file and sparse2d_to_file methods.

lib/include/tick/array/sparsearray2d.h
Mbompr: A cpp unit test

Dekken: It's a bit more complicated that that as it requires python to be run first but yeah can be

lib/include/tick/array/sparsearray2d.h
Dekken: The save/load functions for SparseArray2d are global and not member functions of the class SparseArray2d, which would not compile.

lib/include/tick/array/sparsearray2d.h
Dekken: sure

tools/python/serializer.py
Dekken: sure

tools/travis/osx_run.sh
Dekken: will be, travis is slow so I was testing multiple things at once

tools/travis/docker_run.sh
Dekken: sure, I think it's useful for debugging tho

lib/include/tick/array/sparsearray2d.h
Mbompr: We can probably make this run in C++ directly by using this sample matrix
```
   _data  = [ 5 8 3 6 ]
   _row_indices = [ 0 0 2 3 4 ]
   _indices = [ 0 1 2 1 ]
```
(from https://en.wikipedia.org/wiki/Sparse_matrix#Compressed_sparse_row_.28CSR.2C_CRS_or_Yale_format.29)

lib/include/tick/array/sparsearray2d.h
Mbompr: Hum, I still don't get it sorry :). They won't compile because they are global and not member, or they need to be global and not member because otherwise they won't compile?
There is probably something I ignore about this compilation problem

lib/include/tick/array/serializer.h
Dekken: I'm not a fan of global public functions but I guess for swig it makes sense.

lib/include/tick/array/sparsearray2d.h
Dekken: They need to be global, which is consistent with array.h save/load methods.

Mbompr: Oh ok, then maybe can we cahnge comment to
> save/load methods need to be global to be consistent with array.h save/load methods.

tools/travis/docker_run.sh
Mbompr: Ok, as you told me we better keep those :)

tools/travis/osx_run.sh
Mbompr: Ok!

lib/src/cpp-test/array/serializer.cpp
```diff
@@ -0,0 +1,20 @@
+
```
Mbompr: Can we make a gtest of it?

lib/src/swig/array/serializer.i
```diff
@@ -0,0 +1,19 @@
...
+
+%include <std_string.i>
+%include <std_shared_ptr.i>
+
+%{
+#include "tick/array/serializer.h"
+%}
+
+// We need to expose this empty class to Python, so that we can pass
+// shared pointers of it to models
```
Mbompr: Deprecated comment :)

setup.py
```diff
@@ -283,7 +283,7 @@ def add_dir_name(dir_name, filenames):
                               '-Ilib/include',
```
Mbompr: Could we avoid to modify setup.py in this PR which is already rich enough

sh/configure_env.sh
```diff
@@ -19,7 +19,7 @@
 
 set -e
 
-TICK_CONFIGURED=1
+echo "Entering configure_env.sh"
```
Mbompr: Should we leave this?

tick/array/tests/serializer.py
```diff
@@ -0,0 +1,35 @@
...
+    tick_double_sparse2d_to_file(features, X)
+    
+    y1 = tick_double_array_from_file(lables)
+    X1 = tick_double_sparse2d_from_file(features)
+    
+    print("X1 type: " + str(type(X1)))
+    print(X1.toarray())
+
+if __name__ == '__main__':
+  unittest.main()
```
Mbompr: You can use `np.test.assert_array_equal` to check array equality

tick/array_test/tests/array_memory_test.py
```diff
@@ -2,12 +2,29 @@
...
+
+def dump_garbage():
+
+    gc.collect()
+    print("\nGARBAGE:")
+    print("\nGARBAGE OBJECTS:")
+    for x in gc.garbage:
+        s = str(x)
+        if len(s) > 80: s = s[:80]
+        print(type(x),"\n  ", s)
```
Mbompr: Tests should not print :)

tick/dataset/serializer.py
```diff
@@ -0,0 +1,32 @@
+# License: BSD 3 clause
```
Mbompr: This could be added as a utility to the dataset module

lib/src/cpp-test/array/serializer.cpp
Dekken: Not strictly, it would require a pre executed python step to create the cereal binary file.

lib/include/tick/array/sparsearray2d.h
```diff
@@ -89,6 +102,33 @@ class SparseArray2d : public BaseArray2d<T> {
...
+
+    static SparseArray2d<T> FROM_FILE(const std::string& _file){
+      SparseArray2d<T> spd2d;
+      {
+        std::ifstream data(_file, std::ios::in | std::ios::binary);
+        cereal::PortableBinaryInputArchive iarchive(data);
+        iarchive(spd2d);
+      }
+      return spd2d;
+    }
```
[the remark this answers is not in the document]
Dekken: I've been using it from C++, but you're right that it's not needed for python

sh/configure_env.sh
Dekken: It was moved to the end of the file

tick/array/tests/serializer.py
Dekken: That function is never called, it was for debugging GC while trying to figure out if SparseArray2d was being deleted properly. Can remove though,

tick/array_test/tests/array_memory_test.py
Dekken: That function is never called, it was for debugging GC while trying to figure out if SparseArray2d was being deleted properly. Can remove though,

tick/dataset/serializer.py
Dekken: Is it not?

setup.py
Dekken: Sure, do you plan on fixing cpplint as part of flattening? :P

tick/array/tests/serializer.py
Dekken: k

PR #140, opened 2017-12-01
Moved all prox and solver source code

stephanegaiffas:
Second iteration for the reorganization and flattening of tick, see #124 
- Moved all `prox` and `solver` python modules and C++ source code
- Updated config files for compilation with mkn
- Updated and reorganized documentation


tick/optim/prox/build/__init__.py
```diff
@@ -1 +0,0 @@
-# License: BSD 3 clause
```
Mbompr: This file must be kept for pip distribution (and moved to ` tick/prox/build/__init__.py`)

tick/optim/solver/build/__init__.py
```diff
@@ -1 +0,0 @@
-# License: BSD 3 clause
```
Mbompr: This file must be kept for pip distribution (and moved to ` tick/solver/build/__init__.py`)

Mbompr:
Without these files I get 
`ImportError: No module named 'tick.prox.build.prox'` error as well.

I think you have these files on your computer but you need to tell git to track them manually as they are ignored. Either by typing `git add -f tick/prox/build/__init__.py` either by tweaking .gitignore file, this would be cleaner I don't know how to set a general rule for all `__init__.py` under `build` folders.

Dekken:
Is this superseded  by #143 ?

stephanegaiffas:
This is superseded by #145

Issue #141, opened 2017-12-04
Fix mac builds on Travis


Dekken:
Since this weekend something has changed to result in travis builds failing within the "travis" part of the code. We don't have access to this so the only real solution is to add

```
set +e
```

Dekken:
Also pyenv upgrade is failing

Mbompr:
Will travis job still fail if we have an error with cpptest / doc / cpplint / python tests ?

stephanegaiffas:
That’s be annoying if not :(


Le 4 décembre 2017 à 13:13:27, Martin (notifications@github.com) a écrit:

Will travis job still fail if we have an error with cpptest / doc / cpplint
/ python tests ?

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/X-DataInitiative/tick/issues/141#issuecomment-348945299>,
or mute the thread

PR #142, opened 2017-12-04
fix macos builds on travis

Dekken:
#141

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,16 +32,22 @@ matrix:
     sudo: required
     env: 
       - PYVER=3.4.6
+    osx_image: xcode9.1
   - os: osx
     language: generic
     sudo: required
     env:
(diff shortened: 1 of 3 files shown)


Dekken:
I hope so, as the change is in the "install" script and not the "run" script

stephanegaiffas:
Okay, but we need to ensure that it’s indeed the case… we need to be sure
that

* python test
* C++ test
* doc (sphinx)
* cpplint

Are indeed tested, otherwise we’ll die of a slow a painful death :)


Le 4 décembre 2017 à 13:31:43, ♦♣♠♥ (notifications@github.com) a écrit:

I hope so, as the change is in the "install" script and not the "run" script

—
You are receiving this because your review was requested.
<https://github.com/X-DataInitiative/tick/pull/142#issuecomment-348949234>,
or mute the thread

PR #147, opened 2017-12-22
Implement Hawkes Cumulant non parametric learner

Mbompr:
Tests might fail on a different platform, but the core of the PR is here.


Mbompr:
Doc of the class
4e0f0207-e1c4-450b-bfeb-836e12a940d6

Generated example
screenshot-2017-12-22 fit hawkes kernel norms using cumulants tick 0 3 0 documentation

Mbompr:
>I really think that we should call this NPHC and not HawkesCumulant since this is how we've called it.

I don't think the same way, it would be less consistent with all the other Hawkes learner names (HawkesExpKern, HawkesSumExpKern, HawkesEM, HawkesADM4, HawkesBasisKernels, HawkesSumGaussians, HawkesConditionalLaw). It would be the only learner that has a non-explicit name. I would agree if it was as known as SVM but in this case, I think it would make it harder to remember its name and understand quickly what this class is used for.

>We should also mention the fact that it requires tensor flow in the documentation of NPHC (didn't see it)

You're right, it should be more explicit

stephanegaiffas:
Alright, let’s go with HawkesCumulantMatching then !

achab:
Good job for the implementation !

Few remarks about the doc:
  - the integration parameter H is not related to the kernel support, so the name "kernel_support" might be a little bit confusing. Something like "integration_support" or "cumulant_truncation" might be better.
  - on the Attributes, n_nodes and dim should have the same name.
  - you could replace "Covariance" and "Skewness" with "Integrated Covariance" and "Integrated Skewness".

Thanks again for the amazing job !

Mbompr:
>the integration parameter H is not related to the kernel support, so the name "kernel_support" might be a little bit confusing. Something like "integration_support" or "cumulant_truncation" might be better.

Hum, I would like this to be as easy to understand as possible. Isn't it intimately linked to the kernel support? I like kernel support because this is was is used by other non-parametric Hawkes estimators. If it behaves the same way I'd like to keep this name and improve the following note:
```
    The kernel support is rather denoted by half with :math:`H` in the
    original paper. It is the quantity for which the truncation of the
    domain of integration to :math:`[-H, H]` for the estimated cumulants
    leads to a small error.
```
Otherwise we can rename it `integration_support` but then we should explain the similarities and differences with `kernel_support`.

>you could replace "Covariance" and "Skewness" with "Integrated Covariance" and "Integrated Skewness".

Is it ok if I only replace this in the doc? Something like
```
    covariance : list of `np.array` shape=(n_nodes,n_nodes)
        Estimated integrated covariance, named :math:`\\widehat{C}` in the paper

    skewness : list of `np.array` shape=(n_nodes,n_nodes)
        Estimated integrated skewness (sliced), named :math:`\\widehat{K^c}` in the paper
```

achab:
The relation between the kernel support and the integration support is not straightforward. A kernel function with compact support triggers a covariance density with a non-finite support. Plus, a kernel's support only takes positive values, while the integration support accepts both positive and negative ones. To avoid any misunderstanding, I would prefer the name `integration_support`.

The parameter `integration_support` controls the maximal lag (positive or negative) upon which we integrate the cumulant densities (covariance and skewness), this amounts to neglect border effects. In practice, this is a good approximation if the support of the kernels is smaller than `integration support` and if the spectral norm of the kernel norms is sufficiently distant from the critical value, namely 1.

Mbompr:
@stephanegaiffas @achab I've made the changes.

@Dekken would you know how I could have travis to install `tensorflow` pip package even if it is not in the lib requirements? It is a quite heavy dependency that is used for this algorithm only so I don't think it should be in the requirements but Travis needs it to run the tests though.

Dekken:
@Mbompr I guess you could edit the travis scripts to install it manually?

stephanegaiffas:
@Mbompr 
Two comments : 
* doc needs to be refactored and put in the tick.hawkes documentation...
* I'm okay to merge this even with no unittests for now, if it's too painful to handle travis...

Mbompr:
@achab , @stephanegaiffas tests have passed, waiting for review update before merging :)

examples/plot_hawkes_cumulants_matching.py
```diff
@@ -0,0 +1,55 @@
+"""
+=======================================
+Fit Hawkes kernel norms using cumulants
+=======================================
+
+This non parametric Hawkes cumulants matching
+(`tick.inference.HawkesCumulantMatching`) algorithm estimates directly
```
stephanegaiffas: `tick.hawkes.HawkesCumulantMatching`

tools/travis/docker_run.sh
```diff
@@ -11,6 +11,7 @@ eval "$(pyenv init -)"
 pyenv global ${PYVER}
 pyenv local ${PYVER}
 
+python -m pip install tensorflow
```
stephanegaiffas: Because of this, can we check that builds are not slowed down a lot ? (just looking at the timing of the builds on travis). If it takes a lot of time, I'd think that we could skip the unittests for this PR

stephanegaiffas:
@Mbompr : I approved this, only minor comments easy to fix, and some general comments

examples/plot_hawkes_cumulants_matching.py
Mbompr: Damn... I'll force push with this and merge directly.

You want an example on real data? Because I have already provided an example on simulated data.

stephanegaiffas:
Oh if there’s an example on simulated data it’s enough.
You can merge no problem. The pip install tensor flow does not slow down
the builds right ?


Le 23 janvier 2018 à 16:39:08, Martin (notifications@github.com) a écrit:

*@Mbompr* commented on this pull request.
------------------------------

In examples/plot_hawkes_cumulants_matching.py
<https://github.com/X-DataInitiative/tick/pull/147#discussion_r163281634>:

> @@ -0,0 +1,55 @@
+"""
+=======================================
+Fit Hawkes kernel norms using cumulants
+=======================================
+
+This non parametric Hawkes cumulants matching
+(`tick.inference.HawkesCumulantMatching`) algorithm estimates directly

Damn... I'll force push with this and merge directly.

You want an example on real data? Because I have already provided an
example on simulated data.

—
You are receiving this because you were mentioned.
<https://github.com/X-DataInitiative/tick/pull/147#discussion_r163281634>,
or mute the thread

tools/travis/docker_run.sh
Mbompr: This build was quicker than the previous build on master... What is really taking time is waiting for an OSX machine to be ready...

achab:
I approve the changes as well ! :)

PR #159, opened 2018-02-01
#152 Templates additions prox/models/solvers for any in/direct base t…

Dekken:
…o add Float to SAGA

### DIFF
--- a/lib/cpp-test/linear_model/models_gtest.cpp
+++ b/lib/cpp-test/linear_model/models_gtest.cpp
@@ -11,12 +11,23 @@
 
 #include "tick/array/array.h"
 #include "tick/linear_model/model_linreg.h"
+#include "tick/robust/model_generalized_linear_with_intercepts.h"
 
 #include <cereal/types/unordered_map.hpp>
 #include <cereal/types/memory.hpp>
 #include <cereal/archives/json.hpp>
 #include <cereal/archives/binary.hpp>
 #include <fstream>
+
+TEST(Model, NoAbstractClasses) {
+  ArrayDouble2d r_features;
+  ArrayDouble r_labels;
+  auto features(r_features.as_sarray2d_ptr());
+  auto labels(r_labels.as_sarray_ptr());
+  TModelLinReg<double, double> modelLinRegdd(features, labels, 0, 1);
+  TModelGeneralizedLinearWithIntercepts<double, double> modelGeneralizedLinearWithInterceptsdd(features, labels, 0, 1);
+  TModelGeneralizedLinear<double, double> modelGeneralizedLinear(features, labels, 0, 1);
+}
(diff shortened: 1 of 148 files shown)


lib/cpp/base_model/model_generalized_linear.cpp
```diff
@@ -2,18 +2,32 @@
 
 #include "tick/base_model/model_generalized_linear.h"
 
-ModelGeneralizedLinear::ModelGeneralizedLinear(const SBaseArrayDouble2dPtr features,
-                                               const SArrayDoublePtr labels,
-                                               const bool fit_intercept,
-                                               const int n_threads)
-  : ModelLabelsFeatures(features, labels),
-    n_threads(n_threads >= 1 ? n_threads : std::thread::hardware_concurrency()),
+template <class T, class K>
```
Mbompr: Do we need 2 templates? Or is this only for atomics? 
Also; K seems to be the "main" template and T only used if we have atomic. I find this confusing and would rather have T as "main" template and K set by default to T (if that's possible).

lib/cpp/base_model/model_generalized_linear.cpp
```diff
@@ -26,20 +40,23 @@ const char *ModelGeneralizedLinear::get_class_name() const {
   return "ModelGeneralizedLinear";
 }
 
-
-double ModelGeneralizedLinear::grad_i_factor(const ulong i,
-                                             const ArrayDouble &coeffs) {
+template <class T, class K>
+K
+TModelGeneralizedLinear<T, K>::grad_i_factor(const ulong i,
+                                             const Array<T> &coeffs) {
```
Mbompr: Dummy question: is there a way to somehow "cast" Array<atomic double> into Array<double> if this array will be read-only? Because this is what will happen in most mdoels: coeffs will be read to compute gradient / loss but a model itself won't write on coeffs.

lib/cpp-test/linear_model/models_gtest.cpp
```diff
@@ -34,7 +45,7 @@ TEST(Model, PartialVsFull) {
   SArrayDoublePtr labels = y.as_sarray_ptr();
   SArrayDouble2dPtr features = x.as_sarray2d_ptr();
 
-  ModelLinReg model(features, labels, false, 2);
+  TModelLinReg<double, double> model(features, labels, false, 2);
```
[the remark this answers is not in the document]
Mbompr: It seems that this is what you have done with `ModelLabelsFeaturesDouble`

stephanegaiffas:
I fully agree on that.... that’s annoying but it’s important to do things incrementally


Envoyé de mon iPhone

Dekken:
Closing to split in smaller PRs

PR #162, opened 2018-02-05
Templating linear models 

Dekken:
Base models and linear models

### DIFF
--- a/lib/cpp/base_model/model_generalized_linear.cpp
+++ b/lib/cpp/base_model/model_generalized_linear.cpp
@@ -2,18 +2,21 @@
 
 #include "tick/base_model/model_generalized_linear.h"
 
-ModelGeneralizedLinear::ModelGeneralizedLinear(const SBaseArrayDouble2dPtr features,
-                                               const SArrayDoublePtr labels,
-                                               const bool fit_intercept,
-                                               const int n_threads)
-  : ModelLabelsFeatures(features, labels),
-    n_threads(n_threads >= 1 ? n_threads : std::thread::hardware_concurrency()),
-    fit_intercept(fit_intercept),
-    ready_features_norm_sq(false) {}
+template <class T>
+TModelGeneralizedLinear<T>::TModelGeneralizedLinear(
+    const std::shared_ptr<BaseArray2d<T>> features,
+    const std::shared_ptr<SArray<T>> labels, const bool fit_intercept,
+    const int n_threads)
+    : TModelLabelsFeatures<T>(features, labels),
+      fit_intercept(fit_intercept),
+      ready_features_norm_sq(false),
+      n_threads(n_threads >= 1 ? n_threads
+                               : std::thread::hardware_concurrency()) {}
 
-void ModelGeneralizedLinear::compute_features_norm_sq() {
+template <class T>
+void TModelGeneralizedLinear<T>::compute_features_norm_sq() {
   if (!ready_features_norm_sq) {
-    features_norm_sq = ArrayDouble(n_samples);
+    features_norm_sq = Array<T>(n_samples);
     // TODO: How to do it in parallel ? (I'm not sure of how to do it)
     for (ulong i = 0; i < n_samples; ++i) {
       features_norm_sq[i] = view_row(*features, i).norm_sq();
@@ -22,24 +25,23 @@ void ModelGeneralizedLinear::compute_features_norm_sq() {
   }
 }
 
-const char *ModelGeneralizedLinear::get_class_name() const {
-  return "ModelGeneralizedLinear";
-}
-
-
-double ModelGeneralizedLinear::grad_i_factor(const ulong i,
-                                             const ArrayDouble &coeffs) {
+template <class T>
+T TModelGeneralizedLinear<T>::grad_i_factor(const ulong i,
+                                            const Array<T> &coeffs) {
   std::stringstream ss;
   ss << get_class_name() << " does not implement " << __func__;
   throw std::runtime_error(ss.str());
 }
 
-void ModelGeneralizedLinear::compute_grad_i(const ulong i, const ArrayDouble &coeffs,
-                                            ArrayDouble &out, const bool fill) {
-  const BaseArrayDouble x_i = get_features(i);
+template <class T>
+void TModelGeneralizedLinear<T>::compute_grad_i(const ulong i,
+                                                const Array<T> &coeffs,
+                                                Array<T> &out,
+                                                const bool fill) {
+  const BaseArray<T> x_i = get_features(i);
   const double alpha_i = grad_i_factor(i, coeffs);
   if (fit_intercept) {
-    ArrayDouble out_no_interc = view(out, 0, n_features);
+    Array<T> out_no_interc = view(out, 0, n_features);
 
     if (fill) {
       out_no_interc.mult_fill(x_i, alpha_i);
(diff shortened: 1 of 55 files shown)


lib/cpp/base_model/model_generalized_linear.cpp
```diff
@@ -22,24 +27,23 @@ void ModelGeneralizedLinear::compute_features_norm_sq() {
...
 
-const char *ModelGeneralizedLinear::get_class_name() const {
-  return "ModelGeneralizedLinear";
-}
-
-
-double ModelGeneralizedLinear::grad_i_factor(const ulong i,
-                                             const ArrayDouble &coeffs) {
+template <class T>
+T
```
Mbompr: I'm not a big fan of putting return type on the line before the method

lib/cpp/linear_model/model_logreg.cpp
```diff
@@ -6,66 +6,75 @@
...
 #include "tick/linear_model/model_logreg.h"
 
-ModelLogReg::ModelLogReg(const SBaseArrayDouble2dPtr features,
-                         const SArrayDoublePtr labels,
-                         const bool fit_intercept,
-                         const int n_threads)
-    : ModelGeneralizedLinear(features, labels, fit_intercept, n_threads),
-      ModelLipschitz() {}
-
-const char *ModelLogReg::get_class_name() const {
```
Mbompr: get_class_name has disappeared

lib/include/tick/base_model/model_generalized_linear.h
```diff
@@ -8,13 +8,29 @@
...
+  using TModelLabelsFeatures<T>::labels;
+  using TModelLabelsFeatures<T>::n_samples;
+  using TModelLabelsFeatures<T>::get_n_samples;
+  using TModelLabelsFeatures<T>::n_features;
+  using TModelLabelsFeatures<T>::get_n_features;
+  using TModelLabelsFeatures<T>::get_label;
+  using TModelLabelsFeatures<T>::get_features;
+  using TModelLabelsFeatures<T>::is_ready_columns_sparsity;
+
+  bool fit_intercept = 0;
```
Mbompr: tiny: I think false is easier to read than 0.

lib/include/tick/base_model/model_generalized_linear.h
```diff
@@ -24,40 +40,47 @@ class DLL_PUBLIC ModelGeneralizedLinear : public ModelLabelsFeatures {
...
+    const std::shared_ptr<BaseArray2d<T> > features,
+    const std::shared_ptr<SArray<T> > labels,
+    const bool fit_intercept,
+    const int n_threads = 1);
+
+  virtual ~TModelGeneralizedLinear() {}
 
-  const char *get_class_name() const override;
+  virtual const char *get_class_name() const {
+    return clazz.c_str();
```
Mbompr: cool :)

lib/include/tick/base_model/model_lipschitz.h
```diff
@@ -24,14 +28,15 @@ class DLL_PUBLIC ModelLipschitz : public virtual Model {
...
-  double lip_mean, lip_max;
+  T lip_mean, lip_max;
+
+  //! All Lipschitz constants
+  Array<T> lip_consts;
 
  public:
-  ModelLipschitz();
+  TModelLipschitz();
+  virtual ~TModelLipschitz() {}
```
Mbompr: Is this empty destructor needed now? Is it just cleaner to add an empty one?

lib/include/tick/robust/model_absolute_regression.h
```diff
@@ -8,25 +8,49 @@
...
+  using TModelGeneralizedLinear<T>::grad_i_factor;
+
  public:
-  ModelAbsoluteRegression(const SBaseArrayDouble2dPtr features,
-                          const SArrayDoublePtr labels,
-                          const bool fit_intercept,
-                          const int n_threads = 1);
+  TModelAbsoluteRegression(
+    const std::shared_ptr<BaseArray2d<T> > features,
+    const std::shared_ptr<SArray<T> > labels,
```
Mbompr: Is it somehow possible to have a shortcut for `std::shared_ptr<SArray<T> >` such as `SArrayPtr<T>`?

lib/include/tick/robust/model_generalized_linear_with_intercepts.h
```diff
@@ -9,36 +9,65 @@
...
+  using TModelGeneralizedLinear<T>::n_threads;
+
+ public:
+  using TModelGeneralizedLinear<T>::grad_i;
+  using TModelGeneralizedLinear<T>::get_features;
+  using TModelGeneralizedLinear<T>::grad_i_factor;
+
  protected:
   /**
+  using TModelGeneralizedLinear<T>::n_samples;
```
Mbompr: copy paste mistake

lib/include/tick/hawkes/model/base/model_hawkes.h
```diff
@@ -10,7 +10,7 @@
 /** \class ModelHawkes
  * \brief Base class of Hawkes models
  */
-class DLL_PUBLIC ModelHawkes : public Model {
+class DLL_PUBLIC ModelHawkes : public TModel<double> {
```
Mbompr: Isn't default `Model` class already `TModel<double>`?

lib/swig/base_model/model.i
```diff
@@ -7,18 +7,50 @@
...
   virtual double loss(const ArrayDouble& coeffs);
-
   virtual unsigned long get_epoch_size() const;
+  virtual bool is_sparse() const;  
+};
+typedef TModel<double> Model;
+typedef std::shared_ptr<Model> ModelPtr;
 
-  virtual bool is_sparse() const;
+%rename(ModelDouble) TModel<double>;
```
Mbompr: Will we need to duplicate swig code for all model classes?

Mbompr: Or can we do something like
```
%template(ModelLabelsFeatures) TModelLabelsFeatures<double>;
typedef TModelLabelsFeatures<double> ModelLabelsFeatures;
```

?

lib/swig/hawkes/model/base/model_hawkes_list.i
```diff
@@ -5,6 +5,8 @@
 #include "tick/hawkes/model/base/model_hawkes_list.h"
 %}
 
+%include "model.i"
```
Mbompr: Did you left this by mistake or is this necessary now?

Dekken: not yet but will be

lib/swig/base_model/model.i
Dekken: This is tricky - I have found that %tempate only works if you have no signatures with templates

i.e. most constructors have references to templated array types, we get type errors unless we specify the whole class using the typedefs "ArrayDouble" instead of "Array<double>"

lib/include/tick/robust/model_absolute_regression.h
Dekken: A) Not without a macro
B) I don't think it will work in SWIG

lib/include/tick/hawkes/model/base/model_hawkes.h
Dekken: This is a forgotten "git checkout"

lib/include/tick/base_model/model_lipschitz.h
Dekken: virtual destructors are sometimes necessary - and are good practice for base classes

https://stackoverflow.com/a/461224/795574

lib/cpp/linear_model/model_logreg.cpp
Dekken: it's now a getter 

https://github.com/Dekken/tick/blob/template_linear_model/lib/include/tick/linear_model/model_logreg.h#L39

lib/include/tick/robust/model_absolute_regression.h
Dekken: I think I actually tried it but SWIG was complaining about typeerrors

lib/include/tick/base_model/model_lipschitz.h
Mbompr: Thanks!

lib/include/tick/robust/model_absolute_regression.h
Mbompr: Not cool, it might worth to open an issue to deal with this afterward. Not a big deal but it might simplify the code.

lib/swig/base_model/model.i
Mbompr: This is really not a good news, I'm afraid that we will end up with methods only added to double... 
Do you think there is a way to do this with a macro?

Dekken: with proper unit tests for all types I don't see the problem

Issue #90, opened 2017-10-05
Add benchmarks


Mbompr:
Let's add some benchmarks that show how tick is used in practice.
This will be used for code analysis with Intel.

Dekken:
https://github.com/Dekken/tick/tree/benchmark ?

Mbompr:
Yes, we should open a PR from it !

PR #164, opened 2018-02-06
Benchmarks

Mbompr:
lib/cmake/modules/FindMKL.cmake needs to be adapted for linux


Mbompr:
I have isolated my modifications in "Martin's modif" commit but I will squash afterwards

Mbompr:
Solves #90

PR #167, opened 2018-02-08
Make least sq weights computation faster

Mbompr:
This was work done at intel workshop. Almost 4 times faster

Benchmark before optimization
results_before_40

Benchmark after optimization
results_after_40


stephanegaiffas:
Wow the difference is huge ? How come cexp is so slower than exponentials ?

PR #171, opened 2018-02-08
template prox module 

Dekken:
#170


lib/cpp/prox/prox.cpp
```diff
@@ -2,60 +2,61 @@
...
+      TICK_ERROR(get_class_name()
+                 << " of range [" << start << ", " << end
+                 << "] cannot be called on a vector of size " << coeffs.size());
   } else {
     // If no range is given, we use the size of coeffs to get it
     set_start_end(0, coeffs.size());
     // But no range was given (set_start_end set has_range to true)
-    // Still it is mandatory to set `start` and `end` has some prox might need to know if
-    // they have changed from one iteration to another (see ProxSortedL1)
+    // Still it is mandatory to set `start` and `end` has some prox might need
```
stephanegaiffas: as some prox :) (my bad)

lib/cpp/prox/prox_separable.cpp
```diff
@@ -74,61 +77,70 @@ double ProxSeparable::call_single(double x,
...
   if (i >= coeffs.size()) {
-    TICK_ERROR(get_class_name() << "::call_single " << "i= " << i << " while coeffs.size()=" << coeffs.size());
+    TICK_ERROR(get_class_name()
+               << "::call_single "
+               << "i= " << i << " while coeffs.size()=" << coeffs.size());
   } else {
     if (has_range) {
       if ((i >= start) && (i < end)) {
-        out[i] = call_single(coeffs[i], step);
+        set_out_i(out, i, call_single(coeffs[i], step));
```
stephanegaiffas: why do we introduce and use a set_ou_i ? And why don't we use it before ? (such as in void TProxSeparable<T>::call(const Array<T> &coeffs, const Array<T> &step, Array<T> &out, ulong start, ulong end))

lib/cpp/prox/prox_separable.cpp
```diff
@@ -74,61 +77,70 @@ double ProxSeparable::call_single(double x,
...
 }
 
-double ProxSeparable::value_single(double x) const {
+template <class T>
+T TProxSeparable<T>::value_single(T x) const {
   TICK_CLASS_DOES_NOT_IMPLEMENT(get_class_name());
 }
+
+template <class T>
+void TProxSeparable<T>::set_out_i(Array<T> &out, size_t i, T d) const {
```
stephanegaiffas: Same question as above ? Why do we need this ? In prevision of atomics ?

lib/cpp/prox/prox_slope.cpp
```diff
@@ -2,34 +2,36 @@
...
+                          ulong end, bool positive)
+    : TProxSortedL1<T>(strength, WeightsType::bh, start, end, positive) {
   this->false_discovery_rate = false_discovery_rate;
 }
 
-const std::string ProxSlope::get_class_name() const {
-  return "ProxSlope";
+template <class T>
+std::string TProxSlope<T>::get_class_name() const {
+  return "TProxSlope";
```
stephanegaiffas: Ah ah here get_class_name returns a string :) Can you do the same as for models ? Namely remove all of them and get the name in the main parent class using idtype ?

lib/include/tick/prox/prox_slope.h
```diff
@@ -6,30 +6,33 @@
...
 
-  ProxSlope(double strength,
-            double false_discovery_rate,
-            ulong start,
-            ulong end,
-            bool positive);
+  TProxSlope(T strength, T false_discovery_rate, ulong start, ulong end,
+             bool positive);
 
-  const std::string get_class_name() const override;
```
stephanegaiffas: Why do we remove the const here ?

lib/cpp/prox/prox_separable.cpp
Dekken: yes good catch! it's not needed yet

lib/include/tick/prox/prox_slope.h
Dekken: const copy is redundant.

lib/cpp/prox/prox.cpp
Dekken: k I'll update

lib/cpp/prox/prox_slope.cpp
Dekken: can we address all "get_class_name" with a single PR?

lib/swig/solver/sto_solver.i
```diff
@@ -9,6 +9,7 @@
 %}
 
 %include "model.i"
+%include "prox.i"
```
Mbompr: Why do we need to include these here now? Because of templates?

Dekken: swig gives typeerrors without it

Mbompr: Ok, swig black magic!

Issue #175, opened 2018-02-13
partially template survival 


Dekken:
While adding float tests to solvers it was required to add templating to survival

PR #176, opened 2018-02-13
template survival module

Dekken:
#175


Mbompr:
Was it necessary to also template ModelSCCS? I'm a bit afraid that this would lead to very bad conflicts with #158 ...

Dekken:
understandable - yet templating half the module seemed a bit peculiar

we'll see what Maryan thinks, I've not changed any functionality.

MaryanMorel:
You're just using templating to allow multiple datatypes for coeffs in ModelSCCS right? In this case, it seems ok

MaryanMorel:
The rebase won't be that hard regaring ModelSCCS I think

Dekken:
I'll revert SCCS and update it after you have merged to be safe - only need

Issue #180, opened 2018-02-19
Fix windows and add appveyor 


Dekken:

Dekken:
completed by #183

PR #183, opened 2018-02-19
fix windows in appveyor

Dekken:
#180


tick/array/build/__init__.py
```diff
@@ -1 +1,7 @@
 # License: BSD 3 clause
+
+import os, platform
+
+if platform.system() == 'Windows':
+  if os.path.dirname(os.path.realpath(__file__)) not in os.environ["PATH"]:
+    os.environ["PATH"] = os.path.dirname(os.path.realpath(__file__)) + os.pathsep + os.environ["PATH"]
```
Mbompr: Do you think this code can be put in an external function? Or it will be problematic to import?

tick/solver/build/__init__.py
```diff
@@ -1 +1,19 @@
 # License: BSD 3 clause
+
```
Mbompr: Maybe could we add a comment telling why we think this is necessary here.

lib/include/tick/solver/sto_solver.h
```diff
@@ -67,10 +68,20 @@ class TStoSolver {
...
 
  public:
-  explicit TStoSolver(int seed = -1);
-
-  TStoSolver(ulong epoch_size = 0, T tol = 0.,
-             RandType rand_type = RandType::unif, int seed = -1);
+  // inline TStoSolver(int seed = - 1) : seed(seed) {
+  //     set_seed(seed);
+  // permutation_ready = false;
+  // }
```
Mbompr: We can remove it but we shouldn't leave it commented imo

Dekken: done

tick/array/build/__init__.py
Dekken: it's possible - the important part is that "__file__" part

lib/cpp-test/export/runtime.cpp
```diff
@@ -0,0 +1,135 @@
+
```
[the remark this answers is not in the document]
Dekken: yup

tick/solver/build/__init__.py
Dekken: yes

Issue #182, opened 2018-02-19
Failed to import modules like hawkes, linear model within tick 


UAQureshi:
Hi there,
I have successfully installed all the dependencies and the library itself. In python I can only import tick and from tick the base_model. I am unable to import any other module. 
Any help in this regard will be appreciated:
Thank you very much.
Below is the error:

`from tick import hawkes`

> ---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
C:\Program Files (x86)\Microsoft Visual Studio\Shared\Anaconda3_64\lib\site-packages\tick\hawkes\model\build\hawkes_model.py in swig_import_helper()
     13         try:
---> 14             return importlib.import_module(mname)
     15         except ImportError:

C:\Program Files (x86)\Microsoft Visual Studio\Shared\Anaconda3_64\lib\importlib\__init__.py in import_module(name, package)
    125             level += 1
--> 126     return _bootstrap._gcd_import(name[level:], package, level)
    127 

C:\Program Files (x86)\Microsoft Visual Studio\Shared\Anaconda3_64\lib\importlib\_bootstrap.py in _gcd_import(name, package, level)

C:\Program Files (x86)\Microsoft Visual Studio\Shared\Anaconda3_64\lib\importlib\_bootstrap.py in _find_and_load(name, import_)

C:\Program Files (x86)\Microsoft Visual Studio\Shared\Anaconda3_64\lib\importlib\_bootstrap.py in _find_and_load_unlocked(name, import_)

C:\Program Files (x86)\Microsoft Visual Studio\Shared\Anaconda3_64\lib\importlib\_bootstrap.py in _load_unlocked(spec)

C:\Program Files (x86)\Microsoft Visual Studio\Shared\Anaconda3_64\lib\importlib\_bootstrap.py in module_from_spec(spec)

C:\Program Files (x86)\Microsoft Visual Studio\Shared\Anaconda3_64\lib\importlib\_bootstrap_external.py in create_module(self, spec)

C:\Program Files (x86)\Microsoft Visual Studio\Shared\Anaconda3_64\lib\importlib\_bootstrap.py in _call_with_frames_removed(f, *args, **kwds)

ImportError: DLL load failed: The specified module could not be found.

During handling of the above exception, another exception occurred:

ModuleNotFoundError                       Traceback (most recent call last)
<ipython-input-15-3f9064cdd63d> in <module>()
      3 from tick import base_model
      4 
----> 5 from tick import hawkes

[...]

PR #184, opened 2018-02-20
fix hawkes imports symbols on windows

Dekken:
#182 

importing hawkes through pytest works but when hawkes is used manually "base_model" cannot be found

+ adding export symbols to classes as linking errors were noted in C++ shared tests


lib/cpp-test/export/runtime.cpp
```diff
@@ -35,22 +35,30 @@
...
+#include "tick/hawkes/model/model_hawkes_sumexpkern_loglik_single.h"
+
+
+#include "tick/hawkes/simulation/simu_hawkes.h"
+#include "tick/hawkes/simulation/simu_poisson_process.h"
+
+#ifdef _MKN_WITH_MKN_KUL_
 #include "kul/log.hpp"
-#include "kul/signal.hpp"
+#include "kul/signal.hpp"#
```
Mbompr: extra `#` at eol?

tick/array/tests/serialize_test.py
```diff
@@ -5,12 +5,11 @@
 
 import numpy as np
 import scipy
+import scipy.sparse
```
Mbompr: is this import used?

Dekken: yes seems required on windows for some reason

lib/cpp-test/export/runtime.cpp
Dekken: whoops

tick/array/tests/serialize_test.py
Mbompr: Oh, ok maybe should we add a comment. My IDE is likely to remove unused imports and it would be problematic with this one!

Dekken:
maybe there's another way around it - it seems odd that it was required but
it fixed the script so I dunno

tick/array/tests/serialize_test.py
Dekken: maybe there's another way around it - it seems odd that it was required but
it fixed the script so I dunno

Mbompr: It's ok to me if it is a temporary fix but I think we should leave a comment telling that this is unused but still necessary for windows compilation.
If we just add this comment we can skip travis and appveyor.

Dekken: possible mismatch between windows and linux perhaps?

https://stackoverflow.com/questions/8696322/why-does-this-attributeerror-in-python-occur

should be good now anyway

Mbompr: Yes this way it becomes used, perfect and thanks for the link!

PR #189, opened 2018-02-23
Set default step for GLM and l2 penalization to a better value

Mbompr:
Useful for comparison with scikit.


tick/base/learner/learner_glm.py
```diff
@@ -167,11 +167,20 @@ def fit(self, X: object, y: np.array):
         if self.step is None and self.solver in self._solvers_with_step:
             if self.solver in self._solvers_with_linesearch:
                 self._solver_obj.linesearch = True
-            elif self.solver == 'svrg':
+            elif self.solver == 'svrg' or self.solver == 'saga':
+                L = self._model_obj.get_lip_max()
+                if self.penalty == 'l2':
```
stephanegaiffas: What if penalty is elastic-net ?

PR #196, opened 2018-02-26
#195 skip tests requiring tensorflow if not found

Dekken:


Dekken:
I have tested this locally with and without tensorflow - if someone would like to double check either scenario that would be nice thanks - all CI has tensorflow installed so that's only 1 use case.

Issue #79, opened 2017-09-26
Cereal serialization of models


stephanegaiffas:
Some models are not serializable; This needs to change :)

Dekken:
I've made a start to this, just linear_model and robust so far - requires serializing all the proxes too while we're at it

just testing with SAGA so far but will add the others. 

how would you like this? in parts? one PR per module?

Mbompr:
I guess proxs or models for example will be able to be serialized on their own. If so, yes I guess if we can have independant PR for this modules it could be cleaner :).

Dekken:
All but survival and hawkes modules have been serialized in C++/Python

Dekken:
https://github.com/X-DataInitiative/tick/pull/250 can probably close this too

PR #199, opened 2018-03-05
add serialization and comparison functions to models

Dekken:
#79

### DIFF
--- a/lib/cpp/linear_model/model_hinge.cpp
+++ b/lib/cpp/linear_model/model_hinge.cpp
@@ -3,13 +3,6 @@
 #include "tick/linear_model/model_hinge.h"
 
 template <class T>
-TModelHinge<T>::TModelHinge(const std::shared_ptr<BaseArray2d<T> > features,
-                            const std::shared_ptr<SArray<T> > labels,
-                            const bool fit_intercept, const int n_threads)
-    : TModelLabelsFeatures<T>(features, labels),
-      TModelGeneralizedLinear<T>(features, labels, fit_intercept, n_threads) {}
-
-template <class T>
 T TModelHinge<T>::loss_i(const ulong i, const Array<T> &coeffs) {
   const T z = get_label(i) * get_inner_prod(i, coeffs);
   if (z <= 1.) {
@@ -30,5 +23,5 @@ T TModelHinge<T>::grad_i_factor(const ulong i, const Array<T> &coeffs) {
   }
 }
 
-template class TModelHinge<double>;
-template class TModelHinge<float>;
+template class DLL_PUBLIC TModelHinge<double>;
+template class DLL_PUBLIC TModelHinge<float>;
--- a/lib/include/tick/array/abstractarray1d2d.h
+++ b/lib/include/tick/array/abstractarray1d2d.h
@@ -127,6 +127,26 @@ class AbstractArray1d2d {
   virtual void _print_dense() const = 0;
   virtual void _print_sparse() const = 0;
 
+  //! @brief Compare two arrays by value - ignores allocation methodology !)
+  bool compare(const AbstractArray1d2d<T> &that) {
+    bool are_equal =
+        this->_size == that._size && this->_size_sparse == that._size_sparse;
+    if (are_equal && this->_indices && that._indices) {
+      for (size_t i = 0; i < this->_size_sparse; i++) {
+        are_equal = this->_indices[i] == that._indices[i];
+        if (!are_equal) break;
+      }
+    }
+    if (are_equal) {
+      for (size_t i = 0; i < size_data(); i++) {
+        are_equal = this->_data[i] == that._data[i];
+        if (!are_equal) break;
+      }
+    }
+    return are_equal;
+  }
+  bool operator==(const AbstractArray1d2d<T> &that) { return compare(that); }
+
  public:
   //! @brief Fill array with zeros (in case of a sparse array we do not
   //! desallocate the various arrays... so it is a "lazy" but quick init !)
--- a/lib/include/tick/array/array.h
+++ b/lib/include/tick/array/array.h
@@ -45,6 +45,9 @@ class SArray;
  */
 template <typename T>
 class Array : public BaseArray<T> {
+  template <class T1>
+  friend std::ostream &operator<<(std::ostream &, const Array<T1> &);
+
  protected:
   using BaseArray<T>::_size;
   using BaseArray<T>::is_data_allocation_owned;
@@ -407,6 +410,11 @@ Array<T> sort_abs(Array<T> &array, Array<ulong> &index,
   // Sort the copy inplace keeping track of the index
   sorted_array.sort_abs(index, increasing);
   return sorted_array;
+}
+
+template <typename T>
+inline std::ostream &operator<<(std::ostream &s, const Array<T> &p) {
+  return s << typeid(p).name() << "<" << typeid(T).name() << ">";
 }
 
 /**
--- a/lib/include/tick/array/basearray2d.h
+++ b/lib/include/tick/array/basearray2d.h
@@ -15,6 +15,9 @@ class Array2d;
  */
 template <typename T>
 class BaseArray2d : public AbstractArray1d2d<T> {
+  template <class T1>
+  friend std::ostream &operator<<(std::ostream &, const BaseArray2d<T1> &);
+
  protected:
   using AbstractArray1d2d<T>::_size;
   using AbstractArray1d2d<T>::_size_sparse;
@@ -132,6 +135,21 @@ class BaseArray2d : public AbstractArray1d2d<T> {
   // This method is defined in sparsearray2d.h
   Array2d<T> as_array2d();
 
+  //! @brief Compare two arrays by value - ignores allocation methodology !)
+  bool compare(const BaseArray2d<T> &that) {
+    bool are_equal = AbstractArray1d2d<T>::compare(that) &&
+                     (this->_n_rows == that._n_rows) &&
+                     (this->_n_cols == that._n_cols);
+    if (are_equal && this->_row_indices && that._row_indices) {
+      for (size_t i = 0; i < _n_rows + 1; i++) {
+        are_equal = (this->_row_indices[i] == that._row_indices[i]);
+        if (!are_equal) break;
+      }
+    }
+    return are_equal;
+  }
+  bool operator==(const BaseArray2d<T> &that) { return compare(that); }
+
  private:
   std::string type() const {
     return (is_dense() ? "Array2d" : "SparseArray2d");
@@ -216,6 +234,11 @@ void BaseArray2d<T>::_print_dense() const {
 template <typename T>
 void BaseArray2d<T>::_print_sparse() const {
   std::cout << "_print_sparse ... not implemented" << std::endl;
+}
+
+template <typename T>
+inline std::ostream &operator<<(std::ostream &s, const BaseArray2d<T> &p) {
+  return s << typeid(p).name() << "<" << typeid(T).name() << ">";
 }
 
   /////////////////////////////////////////////////////////////////
--- a/lib/include/tick/base/debug.h
+++ b/lib/include/tick/base/debug.h
@@ -40,6 +40,46 @@
 #define TICK_DEBUG_VERBOSE_MODE 1
 #endif
 
+/** MACRO TICK_CMP_REPORT
+* This exists for comparison of two objects of the exact same type
+*  so that we can compare values of variables, and know when they differ
+*  Its a lambda which receives all values as reference
+* an example looks like
+
+  bool compare(const Object<T> &that) {
+    std::stringsteam ss;
+    bool are_equal = TICK_CMP_REPORT(ss, variable_on_class);
+    if(are_equal){
+      std::cerr << "Objects differ \n" << ss.str();
+    }
+    return are_equal;
+  }
+*/
+#define TICK_CMP_REPORT(ss, var)                                       \
+  [&]() {                                                              \
+    bool are_equal = this->var == that.var;                            \
+    if (!are_equal)                                                    \
+      ss << #var ": " << this->var << " != " << that.var << std::endl; \
+    return are_equal;                                                  \
+  }()
+/** TICK_CMP_REPORT_PTR
+ * This is the same as the above macro expect it includes a dereference for
+ *  pointers or std::shared_ptr etc - null checks are included
+ */
+#define TICK_CMP_REPORT_PTR(ss, var)                                      \
+  [&]() {                                                                 \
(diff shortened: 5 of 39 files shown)


lib/include/tick/array/array.h
```diff
@@ -45,6 +45,9 @@ class SArray;
  */
 template <typename T>
 class Array : public BaseArray<T> {
+  template <class T1>
```
Mbompr: Why T1 here and T on line 415? 
Can't this be mixed with `tick::TemporaryLog<E>& operator<<(tick::TemporaryLog<E>& log, const Array<T>& arr)`?

lib/include/tick/array/basearray2d.h
```diff
@@ -132,6 +135,18 @@ class BaseArray2d : public AbstractArray1d2d<T> {
   // This method is defined in sparsearray2d.h
   Array2d<T> as_array2d();
 
+  bool compare(const BaseArray2d<T> &that) {
+    bool b = (this->_n_rows == that._n_rows) && (this->_n_cols == that._n_cols);
```
Mbompr: I would find cleaner to first initialize `b` to `AbstractArray1d2d<T>::compare(that);`, so that we can rely on the fact that we will test next attributes only if the basic ones are already checked.
(and we can change `b` variable name :) )

lib/include/tick/base/debug.h
```diff
@@ -40,6 +40,20 @@
 #define TICK_DEBUG_VERBOSE_MODE 1
 #endif
 
+#define TICK_CMP_REPORT(ss, v)                                         \
```
Mbompr: Can we change variable names :)? And maybe add some comments to remember how it works in 6 months!

lib/include/tick/base/report.h
```diff
@@ -0,0 +1,51 @@
...
+  BoolStrReport& operator=(const BoolStrReport&& r) = delete;
+
+  std::string why() { return str; }
+
+  explicit operator bool() const { return success; }
+};
+
+template <typename T>
+inline std::ostream& operator<<(std::ostream& s, const BoolStrReport& r) {
+  return s << r.str;
```
Mbompr: Can't we use r.why() and avoid friend function? 
I would still love to have `len(variable_name) > 3` :)

lib/include/tick/base/report.h
```diff
@@ -0,0 +1,51 @@
...
+  bool success = false;
+  std::string str;
+
+ public:
+  BoolStrReport(bool b, const std::string& s) : success(b), str(s) {}
+  BoolStrReport(BoolStrReport& b, const std::string& s)
+      : success(b.success), str(s) {}
+  explicit BoolStrReport(const std::pair<bool, std::string>& p)
+      : success(p.first), str(p.second) {}
+  explicit BoolStrReport(const std::pair<BoolStrReport, std::string>& p)
```
Mbompr: With this constructor you completely ignore the `str` attribute of the BoolStrReport, is this wanted?

lib/include/tick/base/report.h
```diff
@@ -0,0 +1,51 @@
...
+#ifndef LIB_INCLUDE_TICK_BASE_REPORT_H_
+#define LIB_INCLUDE_TICK_BASE_REPORT_H_
+
+#include <sstream>
+
+template <class A = bool, class B = std::string>
+class Report {};
+
+class BoolStrReport : public Report<bool, std::string> {
+  template <class T1>
```
Mbompr: What is T1 used for?

lib/include/tick/base_model/model.h
```diff
@@ -10,12 +10,16 @@
...
 #include <cereal/cereal.hpp>
+#include <cereal/types/polymorphic.hpp>
 
 #include <iostream>
 
 // TODO: Model "data" : ModeLabelsFeatures, Model,Model pour les Hawkes
 template <class T>
 class TModel {
+  template <class T1>
+  friend std::ostream &operator<<(std::ostream &, const TModel<T1> &);
```
Mbompr: Why is T1 needed, can't it be T?

lib/include/tick/base/report.h
Mbompr: Oh, I think I understand, you will report the first one which fails? Perhaps this could be told explicitly somewhere :).

lib/include/tick/base_model/model_labels_features.h
```diff
@@ -73,15 +73,40 @@ class DLL_PUBLIC TModelLabelsFeatures : public virtual TModel<T> {
...
     ar(cereal::make_nvp("labels", *labels));
     ar(cereal::make_nvp("features", *features));
   }
+
+ protected:
+  BoolStrReport compare(const TModelLabelsFeatures<T> &that,
+                        std::stringstream &ss) {
+    if (!this->features || !that.features)
+      TICK_ERROR("Cannot compare null features");
+    if (!this->labels || !that.labels) TICK_ERROR("Cannot compare null labels");
```
Mbompr: Is there no way we can encompass this check in the array compare?

lib/include/tick/linear_model/model_hinge.h
```diff
@@ -21,19 +24,38 @@ class DLL_PUBLIC TModelHinge : public virtual TModelGeneralizedLinear<T> {
   using TModelGeneralizedLinear<T>::get_inner_prod;
   using TModelGeneralizedLinear<T>::get_class_name;
 
+ private:
+  TModelHinge() : TModelHinge<T>(nullptr, nullptr, false) {}
```
Mbompr: Can we mention that is used for serialization purposes only?

lib/include/tick/linear_model/model_hinge.h
```diff
@@ -21,19 +24,38 @@ class DLL_PUBLIC TModelHinge : public virtual TModelGeneralizedLinear<T> {
...
   template <class Archive>
   void serialize(Archive &ar) {
-    ar(cereal::make_nvp("ModelGeneralizedLinear",
-                        cereal::base_class<ModelGeneralizedLinear>(this)));
+    ar(cereal::make_nvp(
+        "ModelGeneralizedLinear",
+        typename cereal::base_class<TModelGeneralizedLinear<T> >(this)));
+  }
+
+  BoolStrReport compare(const TModelHinge<T> &that, std::stringstream &ss) {
```
Mbompr: Can't this be generic and shared across models?

lib/include/tick/linear_model/model_poisreg.h
```diff
@@ -11,37 +11,51 @@
 
 // TODO: labels should be a ArrayUInt
 
-enum class LinkType { identity = 0, exponential };
+enum class LinkType : uint16_t { identity = 0, exponential };
```
Mbompr: Is `uint16_t` needed? If not I think it just makes the code more frightening.

lib/cpp/linear_model/model_hinge.cpp
```diff
@@ -2,13 +2,6 @@
 
 #include "tick/linear_model/model_hinge.h"
 
-template <class T>
-TModelHinge<T>::TModelHinge(const std::shared_ptr<BaseArray2d<T> > features,
-                            const std::shared_ptr<SArray<T> > labels,
-                            const bool fit_intercept, const int n_threads)
-    : TModelLabelsFeatures<T>(features, labels),
-      TModelGeneralizedLinear<T>(features, labels, fit_intercept, n_threads) {}
-
```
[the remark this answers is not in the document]
Dekken: necessary for windows, if a class is templated it's kinda supposed to entirely be in a header so yeah

lib/include/tick/array/abstractarray1d2d.h
```diff
@@ -127,6 +127,25 @@ class AbstractArray1d2d {
   virtual void _print_dense() const = 0;
   virtual void _print_sparse() const = 0;
 
+  bool compare(const AbstractArray1d2d<T> &that) {
+    bool b =
```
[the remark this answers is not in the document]
Dekken: are_supers_equal

lib/include/tick/array/abstractarray1d2d.h
```diff
@@ -127,6 +127,25 @@ class AbstractArray1d2d {
...
+        this->_size == that._size && this->_size_sparse == that._size_sparse;
+    if (b && this->_indices && that._indices) {
```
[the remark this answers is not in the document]
Dekken: I don't disagree, to be if(this->_indices) is the same as if(this->_indices == nullptr)

I'll have a look

lib/include/tick/array/abstractarray1d2d.h
```diff
@@ -127,6 +127,25 @@ class AbstractArray1d2d {
...
+    }
+    if (b) {
+      for (size_t i = 0; i < size_data(); i++) {
+        b = this->_data[i] == that._data[i];
+        if (!b) break;
+      }
+    }
+    return b;
+  }
+  bool operator==(const AbstractArray1d2d<T> &that) { return compare(that); }
```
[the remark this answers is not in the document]
Dekken: I was thinking about adding a this == &that  - sure - I was going to add a doxgen ticket separately as I think it's a bit lacking in general

lib/include/tick/array/array.h
Dekken: T1 is required as it's a thing called "template shadowing" 

it's good in this case to let the compiler know that May be a different type(not that will be)

lib/include/tick/base_model/model.h
Dekken: T1 is required as it's a thing called "template shadowing" 

it's good in this case to let the compiler know that May be a different type(not that will be)

lib/include/tick/base_model/model_labels_features.h
Dekken: not really, they're shared_ptrs which might be null - and thus we might have no reference to an array

lib/include/tick/linear_model/model_hinge.h
Dekken: sure

lib/include/tick/linear_model/model_hinge.h
Dekken: Not as we have it no, the types don't exist on a superclass, not without implementing that templating thing I was telling you about.

lib/include/tick/linear_model/model_poisreg.h
Dekken: This is good practice to specify the "underlying" type

[See](https://github.com/X-DataInitiative/tick/pull/199/files/dc5b9b3a14cd61ee368a27472ad29bccfae02902#diff-3279f659de5f919d107b73e16c0dce09R16)

lib/cpp/linear_model/model_hinge.cpp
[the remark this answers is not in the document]
Mbompr: Ok!

lib/include/tick/array/abstractarray1d2d.h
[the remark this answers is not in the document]
Mbompr: Well you can't write `return are_supers_equal` at the end of the method

lib/include/tick/array/array.h
Mbompr: Could it be? Because if it will never be I rather think it makes the code more difficult to understand

lib/include/tick/base_model/model_labels_features.h
Mbompr: Ok, and can't `TICK_CMP_REPORT_PTR` check that if one of them is a nullptr then the other should be as well?

lib/include/tick/linear_model/model_poisreg.h
Mbompr: Alright!

lib/cpp-test/serialization/solver_gtest.cpp
```diff
@@ -0,0 +1,408 @@
...
+#include <cereal/types/unordered_map.hpp>
+
+#include <gtest/gtest.h>
+
+
+#ifdef _MKN_WITH_MKN_KUL_
+#include "kul/log.hpp"
+#include "kul/dbg.hpp"
+#endif
+
```
Mbompr: Maybe should emphasize that this is for test only!

lib/cpp-test/serialization/solver_gtest.cpp
```diff
@@ -0,0 +1,408 @@
...
+#include "tick/prox/prox_l2sq.h"
+#include "tick/prox/prox_multi.h"
+#include "tick/prox/prox_positive.h"
+#include "tick/prox/prox_slope.h"
+#include "tick/prox/prox_sorted_l1.h"
+#include "tick/prox/prox_tv.h"
+
+#include "tick/solver/saga.h"
+
+namespace testing
```
Mbompr: Is this something we would like to share with all tests?

lib/cpp-test/serialization/solver_gtest.cpp
```diff
@@ -0,0 +1,408 @@
...
+::testing::AssertionResult IsTrue(BoolStrReport &&b) {
+  if (!b) return ::testing::AssertionFailure() << b.why();
+  return ::testing::AssertionSuccess();
+}
+
+template <class T, typename InputArchive, typename OutputArchive>
+class Wrapper {
+ public:
+  template <class SOLVER, class MODEL, class PROX>
+  static void TBD(const std::shared_ptr<SOLVER> solver_ptr,
```
Mbompr: TBD?

lib/cpp-test/serialization/solver_gtest.cpp
```diff
@@ -0,0 +1,408 @@
...
+      InputArchive inputArchive(os);
+      inputArchive(re_model);
+
+      ASSERT_TRUE(IsTrue((*model_ptr.get() == re_model)));
+
+      Array<T> out_grad_restored(2);
+      re_model.grad(coeffs, out_grad_restored);
+      ASSERT_TRUE(out_grad == out_grad_restored);
+    }
+  }
```
Mbompr: Does this involve solver and prox or only model?

lib/cpp-test/serialization/solver_gtest.cpp
```diff
@@ -0,0 +1,408 @@
...
+  test_saga_linear_seperable<float, TModelLinReg<float> >([](const SBaseArrayFloat2dPtr &features,
+                                           const SArrayFloatPtr &labels) {
+    return std::make_shared<TModelLinReg<float> >(features, labels, false, 1);
+  });
+}
+//####################### LIN REG ############################################
+
+//####################### LOG REG ############################################
+TEST(Model, LogRegDoubleSerializationJSON) {
+  test_saga_linear_seperable<double, TModelLogReg<double> >(
```
Mbompr: separable?

PR #205, opened 2018-03-07
#204 reintroduce Werror and work around failures in cereal

Dekken:


Mbompr:
Ok if we keep only the Werror commit

Dekken:
ooh I thought I reset this to remove that commit >_>

lib/include/tick/base/serialization.h
```diff
@@ -3,8 +3,18 @@
 
 // License: BSD 3 clause
 
+#include "tick/base/defs.h"
+
+#ifndef TICK_SWIG_INCLUDE
```
Mbompr: Maybe should we just tell that this is a windows warning code :)

Issue #206, opened 2018-03-07
ICC issue with DBL_MAX


Dekken:
Since templating model_coxreg_partial_lik

ICC has an error compiling it

PR #207, opened 2018-03-07
#206 remove DBL_* in templated code and replace with std numeric_limits

Dekken:
#206


Mbompr:
Maybe can we remove it from time_func as well!

PR #210, opened 2018-03-09
Serialise prox

Dekken:
This contains the serializing models commit, which once merged should be removed from the diff of this PR

### DIFF
--- a/lib/include/tick/prox/prox_binarsity.h
+++ b/lib/include/tick/prox/prox_binarsity.h
@@ -7,6 +7,9 @@
 
 template <class T>
 class DLL_PUBLIC TProxBinarsity : public TProxWithGroups<T> {
+  // Grants cereal access to default constructor
+  friend class cereal::access;
+
  protected:
   using TProxWithGroups<T>::proxs;
   using TProxWithGroups<T>::is_synchronized;
@@ -16,6 +19,9 @@ class DLL_PUBLIC TProxBinarsity : public TProxWithGroups<T> {
   using TProxWithGroups<T>::get_class_name;
 
  protected:
+  // This exists soley for cereal which has friend access
+  TProxBinarsity() : TProxBinarsity(0, nullptr, nullptr, false) {}
+
   std::unique_ptr<TProx<T> > build_prox(T strength, ulong start, ulong end,
                                         bool positive) override;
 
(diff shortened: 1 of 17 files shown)


lib/include/tick/prox/prox_binarsity.h
```diff
@@ -27,19 +33,35 @@ class DLL_PUBLIC TProxBinarsity : public TProxWithGroups<T> {
                  SArrayULongPtr blocks_length, ulong start, ulong end,
                  bool positive);
 
-  TProxBinarsity() = delete;
-  TProxBinarsity(const TProxBinarsity& other) = delete;
-  TProxBinarsity(TProxBinarsity&& other) = delete;
-  TProxBinarsity& operator=(const TProxBinarsity& other) = delete;
-  TProxBinarsity& operator=(TProxBinarsity&& other) = delete;
```
Mbompr: Are they unnecessary now? I thought it was best practice to set these.

lib/include/tick/prox/prox_l1w.h
```diff
@@ -49,6 +56,22 @@ class DLL_PUBLIC TProxL1w : public TProxSeparable<T> {
 
   void set_weights(SArrayTPtr weights) { this->weights = weights; }
 
+  template <class Archive>
+  void serialize(Archive &ar) {
+    ar(cereal::make_nvp("ProxSeparable",
+                        cereal::base_class<TProxSeparable<T>>(this)));
+
+    ar(CEREAL_NVP(*weights));
```
Mbompr: Does this work this simply? I would have thought it would have been necessary to do something similar to:
https://github.com/Dekken/tick/blob/master/lib/include/tick/base_model/model_labels_features.h#L64

lib/include/tick/prox/prox_with_groups.h
```diff
@@ -100,6 +101,30 @@ class TProxWithGroups : public TProx<T> {
     this->blocks_length = blocks_length;
     is_synchronized = false;
   }
+
+ protected:
+  template <class Archive>
+  void serialize(Archive &ar) {
+    ar(cereal::make_nvp("Prox", cereal::base_class<TProx<T> >(this)));
+    ar(CEREAL_NVP(positive));
```
Mbompr: `TProxWithGroups` should not have its own `positive` and should be using `TProx` one!

lib/include/tick/prox/prox_binarsity.h
Dekken: IMO it is, but unless we're going to put them everywhere I don't see the point.

I had added them while I was trying to figure out what was wrong with building on windows. it turned out the entire source of ProxWithGroups had to in the header

lib/include/tick/prox/prox_l1w.h
Dekken: I recall that, and perhaps, the gtest seems ok so perhaps we can validate from that?

lib/include/tick/prox/prox_with_groups.h
Dekken: Yeah I thought that was a bit strange

lib/include/tick/prox/prox_binarsity.h
Mbompr: Ok your choice, I wouldn't have even thought to add them :)!

lib/include/tick/prox/prox_with_groups.h
Mbompr: DO we fix this here or in another issue?

lib/include/tick/prox/prox_binarsity.h
Dekken: https://cpppatterns.com/patterns/rule-of-five.html

lib/include/tick/prox/prox_with_groups.h
Dekken: https://github.com/X-DataInitiative/tick/issues/213

lib/include/tick/prox/prox_l1w.h
```diff
@@ -49,6 +56,22 @@ class DLL_PUBLIC TProxL1w : public TProxSeparable<T> {
 
   void set_weights(SArrayTPtr weights) { this->weights = weights; }
 
+  template <class Archive>
+  void serialize(Archive &ar) {
+    ar(cereal::make_nvp("ProxSeparable",
+                        cereal::base_class<TProxSeparable<T>>(this)));
+
+    ar(CEREAL_NVP(weights));
```
Mbompr: So does this one pass the gtests? Is so I think we can merge :)

Dekken: I have tested it on a future branch with solvers and yes it does

Mbompr: Alright, let's merge then !

PR #212, opened 2018-03-12
Serial solver

Dekken:
includes prox serializing from other PR


Dekken:
This is fairly similar to the previous commits - and once merged should make the python serializing diff smaller

appveyor.yml
```diff
@@ -40,35 +40,36 @@ build_script:
...
-  - C:\Python35-x64\Scripts\pip install -r requirements.txt
-  - C:\Python35-x64\Scripts\pip install tensorflow
-  - C:\Python35-x64\python setup.py build_ext --inplace 
-  - C:\Python35-x64\python -m unittest discover -v . "*_test.py"
-  - C:\Python35-x64\python setup.py bdist_wheel
-  - bash -c "PATH=\"$PWD:/c/Python35-x64/:/c/Program Files/Git/usr/bin:$PATH\"; sh/mkn.sh"
-  - bash -c "PATH=\"$PWD:/c/Python35-x64/:/c/Program Files/Git/usr/bin:$PATH\"; sh/gtest.sh" 
-  - rm -rf build
-  - rm -rf lib\bin
+  # # Python 3.5 # with addition of gtests appveyor is taking more than an hour and timing out
```
Mbompr: Is this still true now that we don't launch all possible combinations?

lib/cpp-test/serialization/solver_gtest.cpp
```diff
@@ -0,0 +1,278 @@
...
+  {
+    auto solver = std::make_shared<TSVRG<T> >(n_samples, 0, RandType::unif, 1e3, -1);
+    SCOPED_TRACE("");
+    WITH_PROX(
+      proxl1, proxl1w, proxl2, proxl2sq, proxelas, proxeq, proxtv, proxpos, prox0
+    ); 
+  }
+}
+
+//####################### LIN REG ############################################
```
Mbompr: Maybe can we put this in model_gtest.cpp?

lib/swig/solver/solver_module.i
```diff
@@ -12,11 +12,16 @@
 
 %{
 #include "tick/base_model/model.h"
+#include "tick/solver/enums.h"
```
Mbompr: We proabably can do 
`%include "tick/solver/enums.h"` so we don't need to put all enum in the swig files

lib/cpp-test/serialization/solver_gtest.cpp
Dekken: it's solvers tho - I removed the part from models_gtest.cpp

appveyor.yml
Dekken: it's uncertain at the moment

lib/swig/solver/solver_module.i
Dekken: sure

lib/cpp-test/serialization/solver_gtest.cpp
Dekken: ok I seem to have left it in models_gtest.cpp - this is a much more extensive set of tests though encapsulating the entire solver/model/prox group

lib/swig/solver/solver_module.i
Dekken: It fails for some reason - I remember trying something like this before and not doing it because of problems

appveyor.yml
Dekken: I will return to this in future and assess - we might want to remove the C++ serialization gtests entirely as it's pretty much done in python

PR #215, opened 2018-03-19
Minimize CI by skipping builds if unnecessary

Dekken:
The thought is to use bintray/appveyor as binary stores to skip the build if we don't think it's necessary 

i.e. if only python/docs stuff has changed


Dekken:
This is not finished, have to see about implementing something for appveyor

Dekken:
This now covers both appveyor and travis

PR #221, opened 2018-04-03
Atomics template

Dekken:
Add second template parameter for atomics

Add ASAGA


Dekken:
should be able to replace the explicit template instantiation nightmare with the following 

```


#include <type_traits>
#include <atomic>

template <class T, class K = T>
class AA{
  K a;
  K b;

 public:
  template <class U = K> typename std::enable_if<std::is_same<U, T>::value, T>::type
  add(){ return a + b;  }

  template <class U = K> typename std::enable_if<std::is_same<U,std::atomic<T>>::value, T>::type
  add(){ return a.load() + b.load();  }
};

int main(int argc, char* argv[]){

    AA<int> a1;
    a1.add();

    AA<int, std::atomic<int>> a2;
    a2.add();

    AA<double> b1;
    b1.add();

    AA<double, std::atomic<double> > b2;
    b2.add();

    return 0;
}
```

Dekken:
Currenly ASAGA uses an inlined version of ProxElasticnet as to most efficiently use atomics without modifying any Proxes - this is something to be analyzed and discussed as there are a number of possibly unnecessary checks which slow down atomic operations 

a possibility is to have safe/unsafe functions, the current ones being the safe kind, and the unsafe ones being where we validate before hand as to avoid checking more often

Dekken:
Recreated as #243

PR #225, opened 2018-04-08
Minimize CI 

Dekken:
by skipping C++ build if only python changes


Dekken:
outstanding: update appveyor builds to push to bintray in the same way as travis

Issue #224, opened 2018-04-07
make doctest has errors


Dekken:
See on master build on Travis

```
plot_conv_sccs_cv_results.py is not compiling:
Traceback (most recent call last):
  File "/Users/travis/build/X-DataInitiative/tick/doc/sphinxext/gen_rst.py", line 885, in generate_file_rst
    execfile(os.path.basename(src_file), my_globals)
  File "/Users/travis/build/X-DataInitiative/tick/doc/sphinxext/gen_rst.py", line 47, in execfile
    code = compile(f.read(), filename, 'exec')
  File "plot_conv_sccs_cv_results.py", line 33
    sim_effects = [*null_effect, intermediate_effect, late_effects]
```

[link](https://travis-ci.org/X-DataInitiative/tick/jobs/359877602)

MaryanMorel:
This bug is caused by the absence of the `*` operator in py 3.4. This operator was introduced in 3.5 (see [pep 448](https://www.python.org/dev/peps/pep-0448/)).

MaryanMorel:
Should be fixed in PR #231

Dekken:
https://github.com/X-DataInitiative/tick/issues/232

Dekken:
resolved by deprecating python 3.4

PR #231, opened 2018-05-03
Bugfix #224

MaryanMorel:
Fix bug due to use of * operator to unpack list, which was introduced only in py 3.5 (see issue #224)


MaryanMorel:
Closed as this bug should be solved by #232

Issue #185, opened 2018-02-20
Add float types to most solvers


Dekken:
Inclusive of unit tests - which may want some additional scrutiny

PR #233, opened 2018-05-07
#185 add float32

Dekken:
Add single template parameter to model/prox/solvers to support "float" type in C++ layer

Add typed class configurations to swig

Add float32 to python layer through dtype inference and "astype" function to return new copies of that type


Mbompr:
Speed comparison example

screenshot-2018-5-24 precision vs speed tradeoff tick 0 4 0 documentation

PR #244, opened 2018-05-25
Addition of CI python formatting checks + format python

Dekken:


Mbompr:
Can we isolate the reformatted .py files from the other in two separated commits?

Dekken:
which "commits"?

PR #246, opened 2018-05-29
Longitudinal preprocessor parallelization

MaryanMorel:
Make the longitudinal preprocessors serializable and make them parallel using python.multiprocessing.

To avoid sending the whole instance of each class to spawned or forked processes, I used an initializer in the process Pool to create a cpp object instance per process. I also used higher-order functions for the same purpose.
The drawback of doing so is having the pool initializers (`_inject_cpp_object` methods) using `global` variables to store the cpp objects. This should not have dire consequences in the multiprocessing context, as each process possesses its own namespace. However, such methods could cause some trouble if called outside of this context (rogue/monkey user).

### DIFF
--- a/tick/preprocessing/longitudinal_features_lagger.py
+++ b/tick/preprocessing/longitudinal_features_lagger.py
@@ -3,10 +3,13 @@
 import numpy as np
 import scipy.sparse as sps
 from tick.preprocessing.base import LongitudinalPreprocessor
-from .build.preprocessing import LongitudinalFeaturesLagger\
+from tick.preprocessing.build.preprocessing import LongitudinalFeaturesLagger\
     as _LongitudinalFeaturesLagger
-from .utils import check_longitudinal_features_consistency,\
+from tick.preprocessing.utils import check_longitudinal_features_consistency,\
     check_censoring_consistency
+from multiprocessing.pool import Pool
+from copy import deepcopy
+from functools import partial, partialmethod
 
 
 class LongitudinalFeaturesLagger(LongitudinalPreprocessor):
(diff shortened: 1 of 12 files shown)


tick/survival/convolutional_sccs.py
```diff
@@ -355,7 +355,8 @@ def fit_kfold_cv(self, features, labels, censoring, C_tv_range: tuple = (),
             features, labels, censoring)
         # split the data with stratified KFold
         kf = StratifiedKFold(n_folds, shuffle, self.random_state)
-        labels_interval = np.nonzero(p_labels)[1]
+        # labels_interval = np.nonzero(p_labels)[1]
```
Dekken: Is there a reason to keep this?

tick/preprocessing/longitudinal_features_lagger.py
```diff
@@ -175,36 +169,57 @@ def transform(self, features, labels=None, censoring=None):
...
+    def _inject_cpp_object(n_intervals, n_lags):
+        """Creates a global instance of the CPP preprocessor object.
+
+        WARNING: to be used only as a multiprocessing.Pool initializer.
+        In multiprocessing context, each process has its own namespace, so using
+        global is not as bad as it seems. Still, it requires to proceed with
+        caution.
+        """
+        global _cpp_preprocessor
+        _cpp_preprocessor = _LongitudinalFeaturesLagger(n_intervals, n_lags)
```
Mbompr: Here is a clever way (using `__call__` in a class) to avoid global variable, did you give it a try?

https://stackoverflow.com/a/37232918/2015762

Dekken:
has this been obsoleted by https://github.com/X-DataInitiative/tick/pull/373 ?

MaryanMorel:
Yep

PR #254, opened 2018-06-18
Atomic solvers

Dekken:


Dekken:
future branch here https://github.com/Dekken/tick/commits/master

lib/cpp/solver/asaga.cpp
```diff
@@ -0,0 +1,224 @@
...
+        //   iterate_j_t
+        // ));
+
+        while (!gradients_average[j].compare_exchange_weak(
+            grad_avg_j,
+            grad_avg_j + (grad_factor_diff * x_ij * n_samples_inverse))) {
+        }
+
+        // Prox is separable, apply regularization on the current coordinate
+        iterate[j] = casted_prox->call_single(
```
Mbompr: Should'nt we use `compare_exchange_weak` here?

Dekken: I don't it's necessary, but feel free to test them

PR #264, opened 2018-07-18
testing docker and python 3.7 on travis

Dekken:


Mbompr:
Why not using ccache anymore?

Dekken:
ccache takes longer on the first build (to build the cache), and it just seemed to not be working as expected

PR #92, opened 2017-10-05
Add python and C++ benchmarks

Mbompr:
Do #90

PR #267, opened 2018-07-24
benchmarks and vector ops

Dekken:
should obsolete #90 #92


Dekken:
replaced by https://github.com/X-DataInitiative/tick/pull/287

PR #280, opened 2018-09-16
Use BASH on travis - format python - was not working

Dekken:
YAPF format checking was not working, as were not using BASH and the checks were failing as a result.


Dekken:
Proof

PR https://travis-ci.org/X-DataInitiative/tick/jobs/429198355#L545
vs 
Master https://travis-ci.org/X-DataInitiative/tick/jobs/427279526#L525

examples/plot_hawkes_conditional_law.py
```diff
@@ -32,10 +32,12 @@
     kernels=[[
         HawkesKernelPowerLaw(multiplier[0], cutoff, exponent, support),
         HawkesKernelPowerLaw(multiplier[1], cutoff, exponent, support)
-    ], [
-        HawkesKernelPowerLaw(multiplier[2], cutoff, exponent, support),
-        HawkesKernelPowerLaw(multiplier[3], cutoff, exponent, support)
-    ]], baseline=[0.05, 0.05], seed=382, verbose=False)
+    ],
```
stephanegaiffas: Why so many files have been badly linted ? (changed indents and such ?

Dekken: This is correct formatting based on your yapf config

Dekken:
We had set travis up to block if python formatting was not correct as per our YAPF config

this error checking was not working as we were not using BASH

Dekken:
https://github.com/X-DataInitiative/tick/blob/master/tools/code_style/yapf.conf

PR #281, opened 2018-09-17
Record history in C++ to allow sto solvers to be quicker

Mbompr:


lib/include/tick/solver/sto_solver.h
```diff
@@ -79,17 +79,33 @@ class DLL_PUBLIC TStoSolver {
   // An array that allows to store the sampled random permutation
   ArrayULong permutation;
 
+  int record_every = 1;
+  int last_record_epoch = 0;
```
Dekken: if it's not a swig value avoid raw int, use size_t or uint64_t / etc

lib/include/tick/solver/sto_solver.h
```diff
@@ -79,17 +79,33 @@ class DLL_PUBLIC TStoSolver {
...
+  double last_record_time = 0;
+
+  // A vector storing all timings at which history has been stored
+  std::vector<double> time_history;
+
+  // A vector storing all epoch at which history has been stored
+  std::vector<int> epoch_history;
```
Dekken: same (uint64_t)

PR #284, opened 2018-09-20
Allow serialization of ModelLabelsFeatureswith sparse features

Dekken:


lib/include/tick/array/array.h
```diff
@@ -526,48 +530,28 @@ CEREAL_LOAD_FUNCTION_NAME(Archive &ar, Array<T> &arr) {
...
+#define ARRAY_DEFINE_TYPE_BASIC(TYPE, NAME)             \
   typedef Array<TYPE> Array##NAME;                      \
   typedef std::vector<Array##NAME> Array##NAME##List1D; \
   typedef std::vector<Array##NAME##List1D> Array##NAME##List2D
 
+#define ARRAY_DEFINE_TYPE(TYPE, NAME)  \
+  ARRAY_DEFINE_TYPE_BASIC(TYPE, NAME); \
+  CEREAL_REGISTER_TYPE(Array##NAME)
+
 ARRAY_DEFINE_TYPE(double, Double);
```
Mbompr: Can we just comment here that we use ARRAY_DEFINE_TYPE_BASIC for all types that don't need serialization yet?

Mbompr: And add that's for compile time reasons

lib/include/tick/array/sparsearray2d.h
```diff
@@ -106,6 +98,103 @@ class SparseArray2d : public BaseArray2d<T> {
...
+
+      this->_data = s_data;
+      this->_indices = s_indices;
+      this->_row_indices = s_row_indices;
+
+      this->is_data_allocation_owned = 1;
+      this->is_indices_allocation_owned = 1;
+      this->is_row_indices_allocation_owned = 1;
+    } catch (const std::exception &e) {
+      std::cerr << e.what() << std::endl;
```
Mbompr: Should we catch this error? Won't this create an "unstable" object?

lib/include/tick/array/sparsearray2d.h
```diff
@@ -106,6 +98,103 @@ class SparseArray2d : public BaseArray2d<T> {
...
+      ar(this->_size);
+
+      inner_save(ar);
+
+      std::vector<INDICE_TYPE> v_indices(this->_indices, this->_indices + this->_size_sparse),
+          v_row_indices(this->_row_indices, this->_row_indices + (this->_n_rows + 1));
+      ar(v_indices);
+      ar(v_row_indices);
+    } catch (const std::exception &e) {
+      std::cerr << e.what() << std::endl;
```
Mbompr: same here

lib/include/tick/array/abstractarray1d2d.h
```diff
@@ -14,14 +14,32 @@
...
+#include <memory>
 #include <type_traits>
 #include <typeinfo>
 
+#include "alloc.h"
 #include "promote.h"
 #include "vector_operations.h"
 
+// clang-format off
+// Don't touch this!
```
Mbompr: Why is this necessary already? Cereal headers raise warnings? Maybe it worths noting it here

Dekken: It was necessary before I added "SortIncludes: False" to the clang format rules, re-arranging includes is not recommended sometimes

lib/include/tick/array/sparsearray2d.h
Dekken: we might want to test a bad file tbh - loading is more error prone than saving

lib/include/tick/array/array.h
Dekken: sure,

lib/include/tick/array/abstractarray1d2d.h
Dekken: still needed it seems,

lib/include/tick/array/sparsearray2d.h
Dekken: loading a bad file causes a segfault when the exception is not caught

lib/include/tick/array/sparsearray2d.h
Dekken: an exception on saving is unlikely, but sometimes better safe than sorry

perhaps we may want a "Array.isValid()" or ".isLoaded()" that exists as a flag for this

lib/include/tick/array/sparsearray2d.h
Dekken: perhaps allowing the segfault on load is the best until we decide how to proceed

PR #291, opened 2018-10-02
Set swig to use relative paths and just use a single search path

Dekken:


setup.py
```diff
@@ -289,15 +289,13 @@ def add_dir_name(dir_name, filenames):
...
-    for opts in [swig_opts, extra_compile_args]:
-        opts.extend(["-Isrc/swig/"+extension_name])
 
     libraries = []
     library_dirs = []
     runtime_library_dirs = []
     extra_link_args = []
     define_macros = []
-    extra_include_dirs = ["include"]
+    extra_include_dirs = ["include", "swig"]
```
Mbompr: This avoids using `opts.extend(["-Isrc/swig/"+extension_name])` ?

Dekken: indeed, we don't need to add each module directory as an include any longer

Dekken:
Yes indeed, thanks for that, I should have put it in the first comment!

Mbompr:
Cool, LGTM :)

PR #293, opened 2018-10-03
test gtest/cmake on appveyor

Dekken:


Mbompr:
This is running cpptest from python setup.py instead of cmake?

Dekken:
cpptest runs cmake

Mbompr:
I'm not getting what this does after #292

Dekken:
previously gtests were done with mkn, which meant we built tick twice - with cpptests/cmake we should just be able to build once all going well

PR #295, opened 2018-10-03
Rearrange vector ops to easily support future changes

Dekken:


lib/include/tick/array/vector_operations.h
```diff
@@ -1,112 +1,41 @@
...
 #include <numeric>
-
 #include <algorithm>
-#include <atomic>
 #include <type_traits>
 
 #include "promote.h"
 #include "tick/base/defs.h"
 
+#if defined(TICK_MKL_AVAILABLE) || defined(TICK_USE_MKL)
```
Mbompr: Why not using only `TICK_USE_MKL` ?

lib/include/tick/array/vector_operations.h
```diff
@@ -1,112 +1,41 @@
...
-};
-
-}  // namespace detail
-}  // namespace tick
-
-#if !defined(TICK_CBLAS_AVAILABLE)
+using vector_operations = detail::vector_operations_mkl<T>;
+}
 
+#elif defined(TICK_CBLAS_AVAILABLE) || defined(TICK_USE_CBLAS)
```
Mbompr: Why not only `TICK_USE_CBLAS `?

lib/include/tick/array/vector/cblas.h
```diff
@@ -1,6 +1,24 @@
 
 
-#if defined(TICK_CBLAS_AVAILABLE)
+#ifndef LIB_INCLUDE_TICK_ARRAY_VECTOR_CBLAS_H_
+#define LIB_INCLUDE_TICK_ARRAY_VECTOR_CBLAS_H_
+
```
Mbompr: I would add
`#if defined(TICK_USE_CBLAS)`
Even if it is only included if `TICK_USE_CBLAS` is defined.

lib/include/tick/array/vector/cblas.h
```diff
@@ -1,6 +1,24 @@
```
Mbompr: I like the way you've called mkl ops_mkl, why not ops_blas and ops_unoptimized ?

lib/include/tick/array/vector/ops_mkl.h
```diff
@@ -0,0 +1,13 @@
+
```
Mbompr: Why do we implement the functions in cblas.h but not in ops_mkl.h? 
Should we have a third file that would call the functions `cblas_*` and include it in both files?
But maybe there is a reason for this dissymmetry.

lib/include/tick/array/vector/unoptimized_impl.h
```diff
@@ -1,5 +1,8 @@
```
Mbompr: why not .hpp anymore?

lib/include/tick/array/vector/unoptimized_impl.h
```diff
@@ -1,5 +1,8 @@
 
+#ifndef LIB_INCLUDE_TICK_ARRAY_VECTOR_UNOPTIMIZED_IMPL_H_
+#define LIB_INCLUDE_TICK_ARRAY_VECTOR_UNOPTIMIZED_IMPL_H_
 
+#include <thread>
```
Mbompr: can be removed

lib/include/tick/array/vector/unoptimized_impl.h
Dekken: whoops

Dekken: nothing is .hpp - so consistency

lib/include/tick/array/vector/ops_mkl.h
Dekken: I think I have a simpler idea

lib/CPPLINT.cfg
```diff
@@ -1,3 +1,3 @@
 set noparent
 linelength=120
-filter=-readability/todo,-legal/copyright,-build/include,-runtime/int,-runtime/references,-whitespace/line_length,-build/c++11
+filter=-readability/todo,-legal/copyright,-build/include,-runtime/int,-runtime/references,-whitespace/line_length,-build/c++11,-readability/casting
```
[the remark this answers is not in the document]
Dekken: removed the change from CPPLINT.cfg and did

https://github.com/X-DataInitiative/tick/pull/295/files#diff-fa496b9c04c596f1b057030571ee1dacR9 

instead - was complaining about casting

PR #297, opened 2018-10-07
corrections for deserializing third party CSR matrices

Dekken:
previous changes to sprasearray serialization prevented deserialization of third party files

### DIFF
--- a/lib/include/tick/array/sparsearray2d.h
+++ b/lib/include/tick/array/sparsearray2d.h
@@ -107,26 +107,47 @@ class SparseArray2d : public BaseArray2d<T> {
     ar(this->_size);
 
     inner_save(ar);
-
-    std::vector<INDICE_TYPE> v_indices(this->_indices, this->_indices + this->_size_sparse),
-        v_row_indices(this->_row_indices, this->_row_indices + (this->_n_rows + 1));
-    ar(v_indices);
-    ar(v_row_indices);
   }
 
+  // Atomic data, BinaryInputArchive
   template <class Archive, typename Y = T>
-  typename std::enable_if<std::is_same<Y, std::atomic<K>>::value>::type inner_save(
-      Archive &ar) const {
-    std::vector<K> k_data;
-    for (size_t i = 0; i < this->_size_sparse; i++) k_data.emplace_back(this->_data[i]);
-    ar(k_data);
+  typename std::enable_if<std::is_same<Y, std::atomic<K>>::value
+    && cereal::traits::is_output_serializable<cereal::BinaryData<T>, Archive>::value, void>::type
+  inner_save(Archive &ar) const {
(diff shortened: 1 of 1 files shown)


lib/include/tick/array/sparsearray2d.h
```diff
@@ -107,26 +107,41 @@ class SparseArray2d : public BaseArray2d<T> {
...
+    ar(cereal::binary_data(this->_row_indices, sizeof(INDICE_TYPE) * (this->_n_rows + 1)));
   }
 
   template <class Archive, typename Y = T>
-  typename std::enable_if<std::is_same<Y, std::atomic<K>>::value>::type inner_save(
-      Archive &ar) const {
-    std::vector<K> k_data;
-    for (size_t i = 0; i < this->_size_sparse; i++) k_data.emplace_back(this->_data[i]);
-    ar(k_data);
+  typename std::enable_if<std::is_same<Y, std::atomic<K>>::value
```
Mbompr: How did you end up knowing we need to differenciate binary and non binary cases?

lib/include/tick/array/sparsearray2d.h
```diff
@@ -107,26 +107,41 @@ class SparseArray2d : public BaseArray2d<T> {
...
+    for (size_t i = 0; i < this->_size_sparse; i++) ar(this->_indices[i]);
+    for (size_t i = 0; i < this->_n_rows + 1; i++) ar(this->_row_indices[i]);
   }
 
   template <class Archive, typename Y = T>
-  typename std::enable_if<!std::is_same<Y, std::atomic<K>>::value>::type inner_save(
-      Archive &ar) const {
-    std::vector<T> v_data(this->_data, this->_data + this->_size_sparse);
-    ar(v_data);
+  typename std::enable_if<!std::is_same<Y, std::atomic<K>>::value
```
Mbompr: Wouldn't `std::enable_if<std::is_same<Y, K>` be clearer than `std::enable_if<!std::is_same<Y, std::atomic<K>>::value` ?
Anyway it might worth adding a comment before each function telling in which cases it applies (like Atomic && binary serialization or Non atomic && non binary-seriation). Because it's painful to decrypt these on Monday morning ;)

lib/include/tick/array/sparsearray2d.h
Dekken: this was already the case, which came about once pickle was brought into the equation

pickling uses "JSON" and doesn't support "binary_data" 

I'd like to see if we can decouple JSON stuff from pickling too

lib/include/tick/array/sparsearray2d.h
Dekken: it would and sure

lib/include/tick/array/sparsearray2d.h
Mbompr: Ok! If we can't get rid of this, maybe we should just add it in a comment :)

PR #300, opened 2018-10-08
Remove requirement for cereal JSON archives in Python/etc

Dekken:
We replace the "JSON" Input/Output archives with hexed "PortableBinary"

Hexed because Python can't seem to tolerate strings containing bytes

Should be smaller too


Dekken:
Travis is slow to respond - it's finished https://travis-ci.org/X-DataInitiative/tick/builds/439126281?utm_source=github_status&utm_medium=notification

Issue #288, opened 2018-09-24
Install Errors


nschiraldi:
I'm struggling to install tick on Oracle Linux Server 7.5, with Python 3.6.5. I've looked at https://github.com/X-DataInitiative/tick/issues/273 and and https://github.com/X-DataInitiative/tick/issues/262 but haven't been able to troubleshoot off of the information from either. I have run:

`
git clone https://github.com/X-DataInitiative/tick.git
git submodule update --init
python3 setup.py build_ext --inplace bdist_wheel
`

which appears to error out with the below. Further, installing via pip causes this import error: ModuleNotFoundError: No module named '_array'

Any help is greatly appreciated.

`creating build/temp.linux-x86_64-3.6/lib/swig/solver
creating build/temp.linux-x86_64-3.6/lib/cpp/solver
compile options: '-Iinclude -Ilib/third_party/cereal/include -I/network/rit/home/ab669414/anaconda3/lib/python3.6/site-packages/numpy/core/include -I/network/rit/home/ab669414/anaconda3/include/python3.6m -c'
extra options: '-D_FILE_OFFSET_BITS=64 -DPYTHON_LINK -DNPY_NO_DEPRECATED_API=NPY_1_7_API_VERSION -Ilib/include -DTICK_SPARSE_INDICES_INT32 -std=c++11 -O2 -DNDEBUG -ffast-math -Wall -Werror -Wno-uninitialized -Isrc/swig/solver -Ilib/swig/array -Ilib/swig/base -Ilib/swig/random -Ilib/swig/base_model -Ilib/swig/linear_model -Ilib/swig/prox -Ilib/swig/robust -DDEBUG_COSTLY_THROW'
gcc: lib/swig/solver/solver_module_wrap.cpp
cc1plus: warning: command line option ‘-Wstrict-prototypes’ is valid for C/ObjC but not for C++ [enabled by default]
gcc: lib/cpp/solver/sgd.cpp
cc1plus: warning: command line option ‘-Wstrict-prototypes’ is valid for C/ObjC but not for C++ [enabled by default]
gcc: lib/cpp/solver/adagrad.cpp
cc1plus: warning: command line option ‘-Wstrict-prototypes’ is valid for C/ObjC but not for C++ [enabled by default]
gcc: lib/cpp/solver/svrg.cpp
cc1plus: warning: command line option ‘-Wstrict-prototypes’ is valid for C/ObjC but not for C++ [enabled by default]
[...]

PR #304, opened 2018-10-12
GCC 4.8 has problems with lambdas, switching to function version

Dekken:
Should resolve https://github.com/X-DataInitiative/tick/issues/288


lib/include/tick/solver/asaga.h
```diff
@@ -55,6 +55,7 @@ class DLL_PUBLIC AtomicSAGA : public TBaseSAGA<T, T> {
   ~AtomicSAGA() {}
 
   void solve(int n_epochs = 1) override;
```
Mbompr: Maybe it worth making it private :)

Mbompr: (wrong line, I meant threaded_solve)

Dekken: done

PR #311, opened 2018-11-07
retry appveyor with python 3.7

Dekken:


Dekken:
still not working

Dekken:
funnily this should work now but we will wait anyway

PR #315, opened 2018-11-21
ModelHawkes.__init__ must be called last 

Dekken:
in init to set default dtype as hawkes models are not templated


Dekken:
I don't think ModelHawkes.__init__ was even called there hence the comment - this is order of operations - but sure can add "must be called last for xyz" .

Mbompr:
Yes, maybe that's a bit different. It's just as this order is not obviously mandatory by simply looking at the code, it worths mentioning that it is done on purpose.

PR #317, opened 2018-12-11
Check for MKL, don't use if not found

Dekken:


Dekken:
seems to be a problem on Travis

tools/python/blas/check_mkl.py
```diff
@@ -0,0 +1,46 @@
...
+
+from setuptools import find_packages, setup, Command
+from setuptools.command.install import install
+from setuptools.extension import Extension
+
+modules = [
+            Extension("mkl", sources=["tools/python/blas/mkl.cpp"], language="c++",)
+          ]
+
+class BLASBuild(build):
```
Mbompr: MKLBuild

tools/python/blas/check_mkl.py
```diff
@@ -0,0 +1,46 @@
...
+    def run(self):
+        self.run_command('build_ext')
+        build.run(self)
+
+setup(name="checkMKL",
+      version='0.5.0.0',
+      ext_modules=modules,
+      install_requires=['scipy',
+                        'numpydoc',
+                        'scikit-learn'],
```
Mbompr: why `install_requires` ?

tools/python/blas/check_mkl.py
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env python
+# -*- coding: utf8 -*-
+
+# python setup.py build_ext --inplace
+
+##
+# This file exists to check if the system
+#  being used to compile Tick has support
+#  for cblas - the "cblas.cpp" file attempts
```
Mbompr: support for mkl ?

tools/python/blas/mkn.cpp
```diff
@@ -0,0 +1,8 @@
+/*
```
Mbompr: Shouldn't this file be called mkl.cpp instead of mkn.cpp ?

setup.py
```diff
@@ -105,6 +105,16 @@
             print("Error executing check_cblas.py - cblas not found")
     elif 'bdist_wheel' not in sys.argv:
         blas_info = get_info("blas_opt", 0)
+    elif platform.system() != 'Linux': # is mac
```
Mbompr: Why don't we check for mkl on linux ?

tools/python/blas/check_mkl.py
Dekken: copy paste

tools/python/blas/check_mkl.py
Dekken: copy pasta

tools/python/blas/mkn.cpp
Dekken: yeep, this happens more often that I'd care to admit

setup.py
Dekken: this issue is mostly about conda on mac, which comes with MKL, but we fail to find the headers

tools/python/blas/check_mkl.py
Dekken: yes

setup.py
```diff
@@ -31,6 +31,11 @@
 from setuptools.command.install import install
 from setuptools.extension import Extension
 
+force_blas = 0
```
Mbompr: Please use True and False :)

PR #331, opened 2019-01-28
Three in one for CI

Dekken:
windows pip updates
dropping 3.7 on CI until more mature - i.e. tensorflow release for 3.7 exists
Numpy Pending Deprecation warning breaking tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,23 +17,25 @@ matrix:
     env:
       - DOCKER_IMAGE=xdatainitiative/tick_ubuntu:1.4.2
       - PYVER=3.6.1
-  - os: linux
-    services: docker
-    env:
-      - DOCKER_IMAGE=xdatainitiative/tick_ubuntu:1.4.2
-      - PYVER=3.7.0
+  ## 3.7 is dropped until more mature
+  # - os: linux
+  #   services: docker
+  #   env:
+  #     - DOCKER_IMAGE=xdatainitiative/tick_ubuntu:1.4.2
+  #     - PYVER=3.7.0
   - os: osx
     language: generic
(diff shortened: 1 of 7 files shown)


tick/array_test/tests/array_memory_test.py
```diff
@@ -259,15 +259,15 @@ def get_memory_used():
...
-                                       delta=1.1 * bytes_size)
+                                       delta=2.1 * bytes_size)
 
                 X = tick_double_sparse2d_from_file(cereal_file)
                 del X
 
             gc.collect()
 
             end = get_memory_used()
-            self.assertAlmostEqual(end, initial_memory, delta=1.1 * bytes_size)
```
Mbompr: Why this high :( ?

Dekken: tests fail

Dekken: we can make a ticket to investigate it

Dekken: https://github.com/X-DataInitiative/tick/issues/332

Mbompr: That sounds frightening... 
It might mean that each time we create an array we have a memory leak...

Dekken: we haven't changed anything, and it doesn't always happen so I dunno what to tell you

Dekken: I was running python 3.5 so let me test some more

Dekken: works on CI now resetting back to 1.1 delta

PR #333, opened 2019-01-30
Werror signed/unsigned comparisons 

Dekken:


Dekken:
This one is pretty straight forward and only happens on some versions of gcc, but it's better to be explicit when possible

PR #336, opened 2019-02-15
Add fetch url dataset helper

Mbompr:


Dekken:
My only comment is to perhaps note that the env vars are not set and so are defaulted.
otherwise looks good

PR #337, opened 2019-02-15
Reorganize benchmarks

Mbompr:
Replace #287, to be merged after #336


setup.py
```diff
@@ -743,6 +743,7 @@ def run(self):
         cmake_cmd = [cmake_exe,
                      '-DPYTHON_INCLUDE_DIR={}'.format(inc_dir),
                      '-DTICK_REBUILD_LIBS=OFF',
+                     '-DBENCHMARK=OFF',
```
Dekken: is this so it can be easily changed to on?  it's "OFF" by default,  it's set to "ON" the benchmark script

Mbompr: Yes, it can be removed as it is off by default but it was to be safe. However, I don't really mind if you think it should be removed.

PR #342, opened 2019-03-11
Main thread hangs when interrupted, exit is required

Dekken:


Dekken:
proof https://github.com/mkn/mkn.kul/blob/master/os/nixish/inc/kul/src/signal/handler.cpp

PR #351, opened 2019-03-27
Faster sccs

Dekken:


lib/cpp/survival/model_sccs.cpp
```diff
@@ -149,8 +154,62 @@ void ModelSCCS::compute_lip_consts() {
...
+
+template <class T, class K>
+void TModelSCCS<T, K>::compute_p_lip_consts(int n_threads) {
+  if (n_threads < 1) TICK_ERROR("ERROR: SCCS n_threads must be positive");
+  unsigned int un_threads = (unsigned int) n_threads;  //  swig can't do uint
+  if (un_threads > n_samples) TICK_ERROR("ERROR: SCCS n_threads must be less than samples");
+  size_t batch = std::ceil(n_samples / n_threads);
+  lip_consts = Array<T>(n_samples);
+  lip_consts.init_to_zero();
+  auto lambda1 = [&](size_t from, size_t to){
```
MaryanMorel: Not a big fan of the naming here (`lambda1`, `i1` a few lines below)

lib/include/tick/survival/model_sccs.h
```diff
@@ -28,27 +36,34 @@ class DLL_PUBLIC ModelSCCS : public ModelLipschitz {
...
 
  public:
-  ModelSCCS() {}  // for cereal
-  ModelSCCS(const SBaseArrayDouble2dPtrList1D &features, const SArrayIntPtrList1D &labels,
-            const SArrayULongPtr censoring, const SArrayULongPtr n_lags);
+  TModelSCCS() {}  // for cereal
+  TModelSCCS(const std::vector<std::shared_ptr<BaseArray2d<T>>>  &features,
+             const SArrayIntPtrList1D &labels,
+             const SArrayULongPtr censoring, const SArrayULongPtr n_lags);
+  // TModelSCCS(const SBaseArray<T>2dPtrList1D &features, const SArrayIntPtrList1D &labels,
```
MaryanMorel: Why do you keep these two commented lines here?

tick/survival/convolutional_sccs.py
```diff
@@ -753,7 +753,7 @@ def _construct_solver_obj(step, max_iter, tol, print_every, record_every,
         # seed cannot be None in SVRG
         solver_obj = SVRG(step=step, max_iter=max_iter, tol=tol,
                           print_every=print_every, record_every=record_every,
-                          verbose=verbose, seed=seed)
+                          verbose=verbose, seed=seed, n_threads=8)
```
MaryanMorel: Why is `n_threads=8` hardcoded here? Maybe use a class attribute to allow the user to change this value.

Btw, Stéphane told me it'd be nice to add the option to use Barzilai Borwein rule, `SVRG(..., step_type='bb')`, which allows running the solver without having to compute the lip const when it's too expensive. It should be user's choice to use either step computed from lip const or using `bb` rule

Dekken:
This PR exists for CI at the moment - comments are premature ;)

Dekken:
clang on MacOS is having this issue https://github.com/swig/swig/pull/1017

can be resolved by
A) upgrading swig to v4
B) compiling with "-Wno-dynamic-class-memaccess"

Dekken:
OSX builds on travis need #356

PR #354, opened 2019-04-16
adding features relatively to the binacox model

SimonBussy:


tick/survival/simu_coxreg.py
```diff
@@ -202,3 +215,274 @@ def _as_dict(self):
...
         return dd
+
+
+class SimuCoxRegWithCutPoints(SimuWithFeatures):
+    """Simulation of a Cox regression for proportional hazards with cut-points
+    effects in the features
+
+    Parameters
+    ----------
+    features : `numpy.ndarray`, shape=(n_samples, n_features), default=`None`
```
stephanegaiffas: There is something that I don't understand. We don't pass coefficients to this simulation class ?

tick/survival/simu_coxreg.py
```diff
@@ -202,3 +215,274 @@ def _as_dict(self):
...
+            candidates = np.percentile(feature_j, quantile_cuts,
+                                       interpolation="nearest")
+            cut_points_j = np.random.choice(candidates, n_cut_points[j],
+                                            replace=False)
+            cut_points_j = np.sort(cut_points_j)
+            cut_points_j = np.insert(cut_points_j, 0, -np.inf)
+            cut_points_j = np.append(cut_points_j, np.inf)
+            cut_points[str(j)] = cut_points_j
+            # generate beta star
+            if j in S:
```
stephanegaiffas: There's no choice in the way we build coefficients in this class ? It's hard coded here ?

tick/survival/simu_coxreg.py
SimonBussy: Indeed: the purpose of the class is to simulate Cox model regression data with cut-points effects in the features. To make sure we get `real` cut-points effects, the later are generated as features quantiles (and hence neither passed to the class), then the coefficients we called `coeffs_binarized` are generated according to the cut-points. In general, the number of coefficients per raw feature are not known in advance. We could also change the class so that we pass the cut-points as well as the binarized coefficients, but with the risk of getting invisible effects (and not answering the aim of the class).

tick/survival/simu_coxreg.py
SimonBussy: Yes, see comment above

Dekken:
we try to use rebasing to keep commits clean

I've done it like so here https://github.com/Dekken/tick/commits/binacox_squash

you can achieve this with "git pull $ORIGIN master --rebase"

Dekken:
this is blocked by https://github.com/X-DataInitiative/tick/pull/366

PR #366, opened 2019-05-27
change comb import to new non deprecated version

Dekken:


MojtabaHayati:
Thanks. How can I do the mentioned replacement?

PR #372, opened 2019-06-20
Array2d Stuff

andro2157:
* Initialize Array2d with std::initializer_list using std::vector

Example :
```
ArrayULong2d array({
  {0,1,0},
  {1,0,1}
});

array = {
  {1,2,3,4}, {5,6,7,8}
};
```

* Convert an Array2d into a SSparseArray2dPtr


Dekken:
The following is now possible

```
  auto random_sparse = SparseArray2d<double>::RANDOM(100, 100, .33);
  auto dense = random_sparse->as_array2d();
  auto sparse = dense.as_sparsearray2d();
```

Dekken:
We might want to consider consistency of types between the conversion functions

one returns an Array2d<T> while the other returns a shared_ptr<SSparseArray2d<T>>

PR #373, opened 2019-06-20
Longitudinal preprocessor multithreading

andro2157:
Multithreading for Longitudinal Features Lagger


lib/include/tick/preprocessing/longitudinal_features_lagger_mp.h
```diff
@@ -0,0 +1,228 @@
+//
```
MaryanMorel: Why don't you put CooMatrix in a separate file?

Dekken: move to ```array/coo_matrix.h```

Dekken:
Should probably have a test of some sort to validate they output the same things

PR #374, opened 2019-06-28
set Swig version required to 4 - provide simple install script

Dekken:


Dekken:
add ```apt-get install -y autotools-dev automake gawk bison flex```

PR #375, opened 2019-06-28
Retrying Python37

Dekken:
Newer versions of python are having issues with the dependency stucture.

Rather than remove Regression/Learner classes from "\_\_init\_\_.py", I have changed the Solver type to a string per Learner, and then import the solvers in the construction function, this way we don't need the solver classes imported until after everything else is loaded properly.


Dekken:
Tensorflow is out for python3.7 - and everything seems ok so

PR #379, opened 2019-07-12
GCC8 Werror changes 

Dekken:
Debian 10 was released the other day, so GCC8 is now the default.


Dekken:
I'm shocked that it ever worked for atomics the way it was

Issue #381, opened 2019-07-21
possible memory leak in python interop for cereal deserialized sparsearray


Dekken:

Dekken:
will see if it's the case for SWIG 3 too if possible

Dekken:
SWIG 3 also seems affected

PR #382, opened 2019-07-21
Force deletion of indices/row_indices in case of deserialized sparsea…

Dekken:
https://github.com/X-DataInitiative/tick/issues/381


Dekken:
I really don't like this, but I've not been able to find out why these are not being deleted either

Issue #390, opened 2019-09-04
HawkesEM error when one of the event arrays is empty


junkyul:
Hi!

I have an error while fitting HawkesEM on the data set possibly has an empty sequence among 2 numpy arrays (D=2 for this data set).
I saw from the previous question that this issue can be resolved by manual install, but the same error is occurring after removing & manual install of tick from source following the instruction. 
When I used HawkesSumExpKern, it worked well.

```
em = HawkesEM(4)
em.fit(data_set)
fig = plot_hawkes_kernels(em, show=False)

for ax in fig.axes:
    ax.set_ylim([0, 1])
plt.show()
```

The following is the error message.

```
  File "/Users/---/hawkes_sumexp.py", line 28, in <module>
    em.fit(data_set)
  File "/Users/---/lib/python3.6/site-packages/tick/hawkes/inference/hawkes_em.py", line 143, in fit
    self.solve(baseline_start=baseline_start, kernel_start=kernel_start)
  File "/Users/---/lib/python3.6/site-packages/tick/solver/base/solver.py", line 109, in solve
    self._solve(*args, **kwargs)
  File "/Users/---/lib/python3.6/site-packages/tick/hawkes/inference/hawkes_em.py", line 184, in _solve
    self._learner.solve(self.baseline, self._flat_kernels)
  File "/Users/---/lib/python3.6/site-packages/tick/hawkes/inference/build/hawkes_inference.py", line 811, in solve
    return _hawkes_inference.HawkesEM_solve(self, mu, kernels)

IndexError: lib/include/tick/array/array.h:106 in T &Array<double>::operator[](const ulong) [T = double]: Bad index: 0 should be in [0, 0]
C++ extension backtrace: 
0   _hawkes_inference.cpython-36m-darwin.so 0x0000001a18abb20b _ZN4tick12TemporaryLogINS_11LogExitNoopEE16insert_backtraceEv + 43
1   _hawkes_inference.cpython-36m-darwin.so 0x0000001a18abb125 _ZN5ArrayIdEixEy + 501
2   _hawkes_inference.cpython-36m-darwin.so 0x0000001a18b3f5ca _ZN8HawkesEM22compute_intensities_urEyRK5ArrayIdER7Array2dId8RowMajorENSt3__18functionIFvdEEEb + 2298
[...]

PR #392, opened 2019-09-04
check if input numpy array is empty in hawkes_em.cpp

junkyul:
One line fix exactly suggested by #390 
no runtime error after passing empty numpy array and passed build_test.sh


Dekken:
hi @junkyul 

would you mind removing the commented out line? then we will merge 

thanks

PR #394, opened 2019-09-09
version number increase

Dekken:


Mbompr:
Do we want to change the version number in files appveyor.yml check_mkl.py and check_cblas.py for consistency?

Dekken:
probably

Dekken:
I''ve updated the manylinux base docker image as the one we were using was causing segfaults for me, plus the github page for it shows a different one

Dekken:
how?

turns out the segfault is a debian thing, and the newer image is too new for auditwheel so I'm running it on ubuntu

Issue #397, opened 2019-09-25
[HawkesExpKern Object Initialization][SegFault]


sara-02:
Initializing the HawkesExpKern Object via pip installed tick, throws a Segmentation Fault Error
Steps to reproduce on Linux with Python 3.6 
Example code taken from [Tick Examples](https://x-datainitiative.github.io/tick/auto_examples/plot_hawkes_matrix_exp_kernels.html#example-plot-hawkes-matrix-exp-kernels-py)
* Install `pip3 install tick`
* `$ python3`
```python
from tick.hawkes import HawkesExpKern
decays = [[4., 1.], [2., 2.]]
learner = HawkesExpKern(decays, penalty='l1', C=10)
```
Error
```
Segmentation fault (core dumped)
```
I tried with float value of decay and got the same error.
As per the gitter conversation regarding this issue, @Dekken confirms that

> **this issue appears to be isolated to the pip binary distribution
it doesn't appear if tick is built from source locally**

Mbompr:
Hi, thanks for reporting, on which platform (pc, mac, linux) do you face this issue ?

sara-02:
@Mbompr Linux, `Ubuntu 18.04.3` to be more specific.

Dekken:
@Mbompr - windows appears fine, if you could confirm Mac is also please.

I'll see about doing a 0.6.0.1 release for this and add dill to setup.py

sara-02:
I just tried the code snippet after building from source on Linux and it worked as expected. Although it did throw a bunch of deprecation warnings while importing HawkesKernExp.

Dekken:
auditwheel folks asked for input as the wheels seem fine before "repairing" https://github.com/pypa/auditwheel/issues/195

PR #399, opened 2019-09-27
import_array needs to be called sooner

Dekken:


Mbompr:
Was this causing the segfault reported in https://github.com/X-DataInitiative/tick/issues/397 ?

Dekken:
import_array is not called early enough

Dekken:
auditwheel repair seems to be the culprit, if that can be assessed this PR is not needed.

Issue #400, opened 2019-09-28
Auditwheel fails not being able to find tick libraries 


Dekken:
setting up LD_LIBRARY_PATH appropriately is the resolution

PR #401, opened 2019-09-29
call import_array statically for linux/mac

Dekken:
- use rpath on linux for auditwheel

fixes https://github.com/X-DataInitiative/tick/issues/400 and https://github.com/X-DataInitiative/tick/issues/397

### DIFF
--- a/lib/swig/tick/array/array_module.i
+++ b/lib/swig/tick/array/array_module.i
@@ -9,9 +9,19 @@
 #include <system_error>
 #include "numpy/arrayobject.h"
 #include "numpy/npy_math.h"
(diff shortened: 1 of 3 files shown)


Dekken:
this resolves the segfault in the "repaired" wheel, but it doesn't resolve the reason for the repair leading to the segfault

MaryanMorel:
Segfault is still creeping around apparently (see Travis logs)

Dekken:
> Segfault is still creeping around apparently (see Travis logs)

I hope this isn't the same thing that is affecting windows with 3.7

PR #409, opened 2019-10-05
block hawkes_cumulant test until tf v2 is supported

Dekken:

PR #407, opened 2019-10-01
keep appveory from updating to tf 2.0

Dekken:

PR #410, opened 2019-10-05
Fix tf version to 1.14 until placeholder is resolved.

sara-02:


Dekken:
this is unnecessary for now

sara-02:
Indeed.

sara-02:
#407, #409 Takes care of it.

PR #463, opened 2020-08-01
some LGTM warning changes

PhilipDeegan:
https://lgtm.com/projects/g/X-DataInitiative/tick/?mode=list

### DIFF
--- a/.github/workflows/build_nix.yml
+++ b/.github/workflows/build_nix.yml
@@ -24,7 +24,7 @@ jobs:
 
     - name: install swig
       run: |
-        git clone https://github.com/swig/swig -b rel-4.0.1 swig
+        git clone https://github.com/swig/swig -b rel-4.0.2 swig
         cd swig && ./autogen.sh && ./configure --without-pcre
         make && make install && cd .. && rm -rf swig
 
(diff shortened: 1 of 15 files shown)


MaryanMorel (approved):

PhilipDeegan:
@MaryanMorel thoughts https://github.com/X-DataInitiative/tick/runs/4990731694?check_suite_focus=true#step:5:1553 ?

MaryanMorel:
> @MaryanMorel thoughts https://github.com/X-DataInitiative/tick/runs/4990731694?check_suite_focus=true#step:5:1553 ?

It seems like the import path is wrong:

```python
from sklearn.metrics.ranking import roc_auc_score
```

should be

```python
from sklearn.metrics import roc_auc_score
```

which is a bit weird, it should have failed a long time ago, unless we were using veeery old versions of scikit-learn maybe?

PhilipDeegan:
> > @MaryanMorel thoughts https://github.com/X-DataInitiative/tick/runs/4990731694?check_suite_focus=true#step:5:1553 ?
> 
> ...


there hasn't been a build in a while. I have fixed that one, but it's more the ``` kf = StratifiedKFold(n_folds, shuffle, self.random_state)``` one that I can't fix

PR #483, opened 2022-01-30
Add newer pythons to github actions

PhilipDeegan:
Lots of things in distutils/setup are being deprecated, so I think by python3.11 it won't work at all. 
Given that I have updated the action builds to use my own build tool [mkn](https://github.com/PhilipDeegan/maiken)

Currently (and previously) on github actions tests we did not run C++ tests as gtest was not installed so it was skipped.
This PR readds C++ tests.

The [googletest](https://github.com/google/googletest) latest code is now using C++14 (or higher) and I have configured mkn to remain tied to the last C++11 version of googletest. I don't see why we couldn't upgrade also to at least C++17.

However:
OSX still has a bunch of errors during pytests for random, which I didn't want to skip for all OSes, more thought is needed here. OSX is still not built in PRs, even though it does build just fails during some pytests.
On Windows, action builds fail with some nefarious DLL init errors, I do not have DLL issues on my local windows install but I will upgrade my compiler to verify as having more versions on windows would be good.
Python3.11 has problem with assertRaisesRegex, so to assess


claudio-ICL (approved):
Looks good to me!

Mbompr (approved):

PR #505, opened 2022-12-14
Hawkes cumulants - Tensorflow v1

claudio-ICL:
This PR proposes to break the class `HawkesCumulantMatching` in a base class with the general interface, and a derived class that implements the methods of the interface using tensorflow. A second derived class that will implement the same methods with pytorch is also prototyped. 
Moreover, this PR tries to migrate `HawkesCumulantMatching` to tensorflow v2 minimising the required changes: I used `tf.compat.v1`. Tests of the inference methodology are re-activated (no longer skipped by default).


Mbompr (approved):

claudio-ICL:
Hi @Mbompr 

I am trying to adapt the code of `HawkesCumulantMatching` to tensorflow v2. There are some tests that fail however, because the numerical precision is not achieved. I was wondering whether you can explain how the expected values were produced?

Mbompr:
If I remember correctly, these values have been obtained by making the algorithm run.
I think that if the new values are close enough, we can simply replace them. Maybe the optimization loop has changed with TF2.

claudio-ICL:
> If I remember correctly, these values have been obtained by making the algorithm run. I think that if the new values are close enough, we can simply replace them. Maybe the optimization loop has changed with TF2.

Understood. So, they are not ground truth. From a first run with v2 I get somewhat similar values, but the precision of 6 decimal places is certainly lost.

image (1)

Mbompr:
I think it makes sense to change the values in the test within the PR then

claudio-ICL:
> > If I remember correctly, these values have been obtained by making the algorithm run. I think that if the new values are close enough, we can simply replace them. Maybe the optimization loop has changed with TF2.
> 
> Understood. So, they are not ground truth. From a first run with v2 I get somewhat similar values, but the precision of 6 decimal places is certainly lost.

I am thinking whether it would be more appropriate to design a test as follows: 
1. Instantiate a parametric class (e.g. exponential) with  given coefficients, and compute exact L1 norms of kernels.
2.  Generate timestamps from simulation.
3. Calibrate `HawkesCumulantMatching` on the simulated timestamps.
4. Compare calibrated `adjacency` with exact L1 norms.

Mbompr:
I agree that makes more sense. But I find it hard to suggest something big when I had been lazy before 😬

tick/hawkes/inference/tests/hawkes_cumulant_matching_test.py
```diff
@@ -8,212 +8,305 @@
...
+from tick.hawkes.inference.hawkes_cumulant_matching import HawkesTheoreticalCumulant
+
+
+class Test(InferenceTest):
+    def setUp(self):
+        self.dim = 2
+        np.random.seed(320982)
+
+    @staticmethod
+    def get_simulated_model():
```
claudio-ICL, line 26: @Mbompr This is the idea that I was suggesting in the conversation of this PR. I simulate a hawkes process with exponential kernels and I train `HawkesCumulantMatching` on five samples. Loading pickled data would be replaced by this, if you agree.
More importantly, the assertion of almost equality now are between the exact theoretical values from the simulated hawkes process and the calibrated coefficients of `HawkesCumulantMatching`. When these assertion succeed, they should guarantee approximate exactness of algorithm.

tick/hawkes/inference/tests/hawkes_cumulant_matching_test.py
```diff
@@ -8,212 +8,305 @@
...
+        if verbose:
+            print('\n')
+            print(f'expected_R_pred:\n{expected_R_pred}')
+            print(f'solution:\n{learner.solution}')
+
+        self.assertTrue(
+            np.allclose(
+                learner.solution,
+                expected_R_pred,
+                atol=0.1,  # TODO: explain why estimation is not so accurate
```
claudio-ICL, line 267: @Mbompr Unfortunately, accuracy is not so good here. Are you familiar with the tuning of hyperparameters of `HawkesCumulantMatching`? If yes, can you please see if a better choice of `step`, `max_iter` etc... gives more accurate estimations? I can see from the paper that the accuracy was really good.

lib/include/tick/hawkes/inference/hawkes_cumulant.h
```diff
@@ -35,4 +33,29 @@ class DLL_PUBLIC HawkesCumulant : public ModelHawkesList {
   }
 };
 
+class DLL_PUBLIC HawkesTheoreticalCumulant {
```
claudio-ICL, line 36: @Mbompr @PhilipDeegan 

This is a new c++ class that implements the theoretical formulae for mean intensity, covariance and skewness of hawkes processes. These are the formulae (7), (8), and (9) in the paper. 
Currently, the class is only used to test the accuracy of the estimation of `HawkesCumulantMatching`. However, the formulae have intrinsic values beside such testing,  and we might include these calculations in a more general Hawkes class in the future.

tick/hawkes/inference/tests/hawkes_cumulant_matching_test.py
claudio-ICL: Hi @Mbompr 

Could you look into this please?

claudio-ICL:
Hi @PhilipDeegan , Do you think we can merge this?

lib/cpp/hawkes/inference/hawkes_cumulant.cpp
```diff
@@ -126,3 +124,66 @@ double HawkesCumulant::compute_E_ijk(ulong r, ulong i, ulong j, ulong k,
   res /= (*end_times)[r];
   return res;
 }
+
+HawkesTheoreticalCumulant::HawkesTheoreticalCumulant(int dim) : d(dim) {
+  Lambda = SArrayDouble::new_ptr(dim);
```
PhilipDeegan, line 129: can these be part of the constuctor setup arguments like `: d(dim)` ?

PhilipDeegan, line 129: 1.  is `Lambda` the best name?
2.  variables names don't typically start with upper case

lib/cpp/hawkes/simulation/simu_hawkes.cpp
```diff
@@ -56,15 +53,13 @@ bool Hawkes::update_time_shift_(double delay, ArrayDouble &intensity,
 void Hawkes::reset() {
   for (unsigned int i = 0; i < n_nodes; i++) {
```
PhilipDeegan, line 54: I know it was `unsigned int` before, I would generally prefer `std::uint32_t` or similar

lib/cpp/hawkes/simulation/simu_hawkes.cpp
```diff
@@ -56,15 +53,13 @@ bool Hawkes::update_time_shift_(double delay, ArrayDouble &intensity,
...
     for (unsigned int j = 0; j < n_nodes; j++) {
-      if (kernels[i * n_nodes + j] != nullptr)
-        kernels[i * n_nodes + j]->rewind();
+      if (kernels[i * n_nodes + j] != nullptr) kernels[i * n_nodes + j]->rewind();
```
PhilipDeegan, line 56: I know it was `unsigned int` before, I would generally prefer `std::uint32_t` or similar

lib/include/tick/hawkes/inference/hawkes_cumulant.h
```diff
@@ -35,4 +33,29 @@ class DLL_PUBLIC HawkesCumulant : public ModelHawkesList {
...
+  SArrayDoublePtr mu;
+  SArrayDoublePtr Lambda;
+  SArrayDouble2dPtr C;
+  SArrayDouble2dPtr Kc;
+  SArrayDouble2dPtr R;
+
+ public:
+  HawkesTheoreticalCumulant(int);
+  int get_dimension() { return d; }
+  void set_baseline(const SArrayDoublePtr mu) { this->mu = mu; }
```
PhilipDeegan, line 48: We might not have it everywhere, but `T const& n` is the preferred order

lib/cpp/hawkes/inference/hawkes_cumulant.cpp
claudio-tw: The variables `Lambda`, `C`, `Kc`, and `R` represent first integrated cumulant, second integrated cumulant, (reduced) third integrated cumulant, and inverse of $I - G$ (where $G$ is the matrix of $L^1$-norms of kernel functions), respectively.  Their names are directly taken from the math symbols used in the [paper](https://arxiv.org/abs/1607.06333): $\Lambda$, $C$, $K$, and $R$, respectively. 

I agree that the upper case is uncommon. I can change the name as follows:
`Lambda` -> `first_cumulant`
`C` -> `second_cumulant`
`Kc` -> `third_cumulant`
`R` -> `g_geometric`

claudio-tw: Done here: 11353f445b59009527ca12464a5eab988bc2b844

lib/cpp/hawkes/simulation/simu_hawkes.cpp
claudio-tw: Understood...
I will open a new PR about that because it is not related to `HawkesCumulant` or tensorflow

lib/include/tick/hawkes/inference/hawkes_cumulant.h
claudio-tw: Changed: fa43f6dc9dc65013c609508c0f4fe43290fe60ac

claudio-ICL:
@PhilipDeegan 
I believe that this PR is ready to merge. Would it be possible to do it today?

PR #508, opened 2023-01-17
HawkesKernel - swig interface issues

claudio-ICL:
Af ew methods of the class `HawkesKernel` give "undefined reference" during the build. How do we fix this?

### DIFF
--- a/lib/swig/tick/hawkes/simulation/hawkes_kernels.i
+++ b/lib/swig/tick/hawkes/simulation/hawkes_kernels.i
@@ -14,6 +14,13 @@ class HawkesKernel {
   double get_primitive_value(double t);
   double get_primitive_value(double s, double t);
   virtual double get_norm(int nsteps = 10000);
+  virtual double get_convolution(
+       const double time, 
+       const ArrayDouble &timestamps, 
+       double *const bound);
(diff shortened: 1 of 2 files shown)


Mbompr:
Does this fix the issue ? Because I cannot find the error message "undefined reference" in the build

lib/swig/tick/hawkes/simulation/hawkes_kernels.i
```diff
@@ -14,6 +14,13 @@ class HawkesKernel {
   double get_primitive_value(double t);
   double get_primitive_value(double s, double t);
   virtual double get_norm(int nsteps = 10000);
+  virtual double get_convolution(
+       const double time, 
+       const ArrayDouble &timestamps, 
+       double *const bound);
+  virtual double get_primitive_convolution(
+       const double time, 
+       const ArrayDouble &timestamps);
```
Mbompr, line 23: This should be needed only if you want to access these functions form Python. Otherwise, it should not be.

claudio-ICL: Hi @Mbompr 

I believe @PhilipDeegan has already resolved the issue. We can close this PR.

Issue #504, opened 2022-12-12
State of Tick in (almost) 2023


PhilipDeegan:
Issues:
- Sparse array serialization segfaults on various platforms with various versions of python
- Various deprecation warnings for csc/csr_matrix
- A number of tests are only failing on OSX

Tests to be disabled generally:
```bash
grep -A 1 -r "\.skip" .
./tick/survival/tests/convolutional_sccs_test.py:    @unittest.skip("issue with StratifiedKFold")
./tick/survival/tests/convolutional_sccs_test.py-    def test_LearnerSCCS_fit_KFold_CV(self):
--
./tick/solver/tests/sdca_test.py:    @unittest.skip("Is failing by a small amount - to assess")
./tick/solver/tests/sdca_test.py-    def test_sdca_identity_poisreg(self):
--
./tick/random/tests/random_test.py:    @unittest.skip("Needs checking with scipy benchmark")
./tick/random/tests/random_test.py-    def test_uniform_int_random(self):
--
./tick/random/tests/random_test.py:    @unittest.skip("Needs some clarity around what we should assert")
./tick/random/tests/random_test.py-    def test_poisson_random(self):
```

For the moment I will be configuring Github actions to only show the builds which do in fact finish. This is not ideal, but considering the state of things I think it's best to have something that we know consistently works for people to make PRs against. Once this is done I can try adding python3.10 on all platforms and checking if it builds ok.

oresthes:
Are there any plans for maintaing this package in the long term?

PhilipDeegan:
@oresthes I'm aiming to get a new release on pypi/pip in the near future

PR #511, opened 2023-02-12
Updated wheel release mechanism

PhilipDeegan:
Biggest change really is now, 
- There is only one C++ library - much simpler for build time

Fixes somethings from https://github.com/X-DataInitiative/tick/issues/504
- no segfaults seen in a while
- csr deprecations assessed

OSX still has some weird issues with random, and I'm guessing this is the reason why some real tests are failing now 

for instance

```
Traceback (most recent call last):
  File "/Users/runner/work/tick/tick/tick/linear_model/tests/logistic_regression_test.py", line 93, in test_LogisticRegression_fit
    self.assertGreater(
AssertionError: 0.6981863527347516 not greater than 0.699 : solver sgd with penalty binarsity and intercept False reached too low AUC
```

I changed it from `.7` to `.69`, but I'm not sure it really *should* be changed, maybe it's ok? 
@Mbompr thoughts?

With this pull request, we should be able to publish new wheels to both test-pypi repo on master merge
And
push to pypi regular on release/tag


Mbompr:
Sorry, just discovering this thread :o, wow the amount of work is huge !

Honestly, these tests involving randomness are very bad designed, but I didn't know that back then :).

About changing the AUC threshold from .7 to .69, this is not an issue at all. The only thing we are checking is that the model is learning something significantly better than random (which would lead to 0.5), so 0.65 or 0.6 would also be fine.

About the darwin test filtering, this is due to

```python
        # This is temporary
        # At the moment, seeds are not cross platform as distributions in
        # C++ depends on standard library shipped with compiler
        import os
        if os.name == 'posix':
            import platform
            if platform.system() == 'Darwin':
                # We check that we get the same as what was recorded
                np.testing.assert_almost_equal(seeded_sample_1, seeded_sample)
```
(note the "temporary"...). That is also bad designed. The test purpose was to check if with the same seed we were obtaining the same random numbers. But nothing really guarantees this and I guess an update in darwin random number generation can break this easily.

I would thus simply remove this part of the test as well as all `seeded_sample` arguments. I would need help tell me, otherwise I'm quite happy to not setup tick, cmake and swig on my machine :)

PR #528, opened 2024-11-27
bugfix: plot_time_function.py

shaisung:
It was plotting the same thing three times (plotting the first term instead of the iterated variable)


Mbompr:
Thank you!

shaisung:
I guess you need to run the build script now :)

PR #532, opened 2026-04-17
Migrate build system and Python bindings to scikit-build-core and pybind11

Mbompr:
## Summary

   - **Replace SWIG + setuptools with pybind11 + scikit-build-core**: the 950-line `setup.py` is replaced by a `pyproject.toml` + `CMakeLists.txt` build, and all `.i` SWIG interface files are rewritten as pybind11 `.cpp` modules
   under `python/src/`. The exposed Python API is unchanged.
   - **Fix cereal serialization for multi-`.so` architecture**: remove cross-SO `CEREAL_REGISTER_TYPE` registrations that break under pybind11's separate shared-library model; split `serialize()` into `load()`/`save()` where cereal
    requires it for non-default-constructible types.
   - **Modernize Python compatibility**: support numpy 2.x (`np.trapezoid`), Python 3.13 (`inspect.signature`), recent sklearn/scipy API changes, and fix several latent C++ bugs (uninitialized members, boundary conditions in
   `TimeFunction`).

   ## Motivation

   The SWIG-based build was fragile, hard to maintain, and blocked support for modern Python (3.12+) and numpy 2.x. scikit-build-core + pybind11 is the current standard for C++ extension packaging, gives cleaner build diagnostics,
   and removes the SWIG dependency entirely.

   ## What changed

   | Area | Before | After |
   |---|---|---|
   | Build system | `setup.py` (setuptools + custom SWIG compilation) | `pyproject.toml` + `CMakeLists.txt` (scikit-build-core) |
   | Python bindings | 60+ SWIG `.i` files under `lib/swig/` | 15 pybind11 `.cpp` files under `python/src/` |
   | Extension loading | Per-module `sys.path` manipulation | Centralised `load_extension()` helper in `tick/base/opsys.py` |
   | CI | macOS CI disabled (`.yml.off`) | All platforms re-enabled and updated |


lib/CMakeLists.txt
```diff
@@ -1,7 +1,13 @@
 cmake_minimum_required(VERSION 3.7)
 project(tick)
 
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -ffast-math")
+set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+set(CMAKE_CXX_EXTENSIONS OFF)
+
+if (NOT MSVC)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -ffast-math")
```
PhilipDeegan: it's not generally advisable to modify "CMAKE_CXX_FLAGS" inside CMakeLists files, the option should be taken at configure time, or an option like

```
option(useFastMath OFF)
```

then you might have a 

```
set(TICK_FLAGS "")
if(useFastMath)
   set(TICK_FLAGS "${TICK_FLAGS} -ffast-math")
endif(useFastMath)
```

then

```
target_compile_options($target $TICK_FLAGS)
```

python/src/tick_base_model_module.cpp
```diff
@@ -0,0 +1,82 @@
...
+#include "tick/base_model/model_labels_features.h"
+#include "tick/base_model/model_lipschitz.h"
+
+namespace py = pybind11;
+
+namespace {
+
+template <typename ModelType, typename Scalar>
+void bind_model_base(py::module_ &m, const char *name) {
+  py::class_<ModelType, std::shared_ptr<ModelType>>(m, name)
```
PhilipDeegan: the "modern" thing to do is like

```
py::class_<ModelType, py::smart_holder>
```

https://pybind11.readthedocs.io/en/stable/advanced/smart_ptrs.html

.github/workflows/build_nix.yml
```diff
@@ -10,18 +10,18 @@ on:
...
 
     strategy:
       fail-fast: false
       max-parallel: 4
       matrix:
-        python-version: ['3.7', '3.9', '3.10', '3.11'] # '3.8' has "'tp_print' is deprecated [-Werror,-Wdeprecated-declarations]"
+        python-version: ['3.11', '3.12', '3.13']
 
     steps:
     - uses: actions/checkout@v2
```
PhilipDeegan: you could probably update all these to v6 while you're at it

lib/mkn.yaml
```diff
@@ -60,7 +60,6 @@ profile:
     self: array
     src:
```
PhilipDeegan, line 61: the only reason I would suggest to keep this (it requires updating similar to how I do it for [statick](https://github.com/PhilipDeegan/statick/blob/master/mkn.yaml))

is that, the CMake approach assumes pybind is installed on the system, whereas mkn will download pybind 

no strong opinion either way you can delete it if you prefer too

PhilipDeegan:
something I had done on another PR was to merge all libraries into one, perhaps you have an argument for keeping them separate tho

Mbompr:
I actually have none, my guiding principle was modernization while keeping as close as possible from master.
The end goal was really that newer Python version are supported and the package becomes easy to install by users again.

.github/workflows/build_nix.yml
```diff
@@ -10,25 +10,25 @@ on:
 env:
   TICK_DEBUG: 0
   TICK_WERROR: 0
```
PhilipDeegan, line 12: these may be redundant now too

PhilipDeegan:
given how long it's been since pypi has been updated, I see nothing wrong with just merging this and then seeing of a patch release needs to be made for manylinux support or whatever

PhilipDeegan (approved):

Mbompr:
Let's give it a try 🔥

